PR #7, opened 2016-11-01
REF: Fully customize docs style

jakereps:
Brings the site to look almost identical to https://qiime2.org with the only difference being the sidebar, since there are nested items it looks wrong being right aligned like the main site.

# Screenshots

Only thing not on this PR is the Analytics tracking.
screen shot 2016-11-01 at 11 49 00 am
screen shot 2016-11-01 at 11 49 05 am


jairideout:
Thanks @jakereps, this is awesome!!!

PR #34, opened 2016-11-30
adds importing sequence data tutorial

### LINKED ISSUE #33: add "importing raw sequence data" tutorial
This should illustrate importing demultiplexed or multiplexed reads, and work from the moving pictures tutorial.

gregcaporaso:
fixes #33 

I debated merging this tutorial with the existing *Importing* tutorial, but I see this one expanding a lot as we support additional input formats so it seems to make more sense on its own. Any objections to that?

### DIFF
--- a/source/tutorials/import-sequence-data.rst
+++ b/source/tutorials/import-sequence-data.rst
@@ -1,0 +1,56 @@
+Importing sequence data
+=======================
+
+.. note:: This guide assumes you have performed the steps in the :doc:`install guide <../install>` and have read :doc:`Importing Data <./import>`.
+
+To begin a QIIME 2 analysis with sequence data, for example as generated on an Illumina MiSeq instrument, it must be imported into a QIIME 2 artifact (``.qza``) file. This document illustrates how to import data in various starting formats, and will be expanded as additional relevant formats are identified. Each section below briefly describes a format of sequence data, provides commands to download example data files, and illustrates how to import data in that format.
+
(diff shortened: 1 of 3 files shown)


jairideout:
I think going with a separate tutorial is a good idea, at least for now. We'll need to think about how to generally (automatically?) document the different semantic types and their supported file formats, as those all apply to importing and exporting. For now, this tutorial will help users move forward and we can always break links and reorganize things in the future since every link to the docs is versioned! 👌

source/tutorials/import-sequence-data.rst
```diff
@@ -0,0 +1,56 @@
...
+   mkdir raw-sequences
+   curl -sL https://data.qiime2.org/2.0.6/tutorials/moving-pictures/raw-sequences/barcodes.fastq.gz > raw-sequences/barcodes.fastq.gz
+   curl -sL https://data.qiime2.org/2.0.6/tutorials/moving-pictures/raw-sequences/sequences.fastq.gz > raw-sequences/sequences.fastq.gz
+
+Importing data
+~~~~~~~~~~~~~~
+
+.. command-block::
+
+   qiime tools import --type RawSequences --input-path raw-sequences/ --output-path raw-sequences
```
gregcaporaso: Thanks. This one actually works if you include the `.qza` extension. That's what we're doing in the *Moving Pictures* tutorial, just made that change here (I think I must have edited it in parallel with testing yesterday, thinking that the extension wasn't necessary).

gregcaporaso:
Thanks for the review @thermokarst! I've addressed all of your comments, build the docs locally, and tested all of the commands by copy/pasting from the built docs. Should be good to go.

PR #43, opened 2016-12-08
ENH: Allow multiline command blocks

### LINKED ISSUE #37: commandblock support for multi-line command execution
It would be great if Sphinx and/or our `commandblock` directive supported multiline commands:

```sh
$ qiime \
  tools \
  import \
  --help
```

jakereps:
Resolves #37

### DIFF
--- a/source/sphinx_directives/command_block.py
+++ b/source/sphinx_directives/command_block.py
@@ -5,6 +5,7 @@ import shutil
 import subprocess
 import tempfile
 import urllib.parse
+import functools
 
 import docutils.nodes
 import docutils.parsers.rst
@@ -36,7 +37,8 @@ class CommandBlockDirective(docutils.parsers.rst.Directive):
 
     def run(self):
         self.assert_has_content()
-        commands = self.content
+        commands = functools.reduce(self._parse_multiline_commands,
+                                    self.content, [])
 
         nodes = [
             self._get_literal_block_node(commands)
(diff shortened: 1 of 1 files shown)


source/sphinx_directives/command_block.py
```diff
@@ -185,6 +187,14 @@ def _get_output_links(self, output_paths, name):
                 content.append('')
         return content
 
+    def _parse_multiline_commands(self, previous, next):
+        if previous and previous[-1].endswith('\\'):
+            previous[-1].replace('\\', '')
+            previous[-1] += next
+        else:
+            previous.append(next)
```
ebolyen: not that it matters, but mutating the accumulator just feel so *wrong* to me...

source/sphinx_directives/command_block.py
```diff
@@ -36,7 +37,8 @@ class CommandBlockDirective(docutils.parsers.rst.Directive):
 
     def run(self):
         self.assert_has_content()
-        commands = self.content
+        commands = functools.reduce(self._parse_multiline_commands,
```
ebolyen: This is a pretty slick solution

source/sphinx_directives/command_block.py
```diff
@@ -185,6 +187,14 @@ def _get_output_links(self, output_paths, name):
                 content.append('')
         return content
 
+    def _parse_multiline_commands(self, previous, next):
+        if previous and previous[-1].endswith('\\'):
+            previous[-1].replace('\\', '')
```
ebolyen: It's unlikely to happen, but just doing a blanket replace might catch things that are escaped. I would refine this to only strip the last character.

jakereps:
Thanks @ebolyen! Got those reducer changes in, and fixed the rendering that I forgot to check before submitting this. Passing the multiline parsed commands into the nodes was breaking the output, so switched it to pass `self.content` directly.

PR #46, opened 2016-12-14
ENH: Add note about cache-busting

jakereps:
Related to: qiime2/cookiecutter-plugin-template/issues/10


jakereps:
Thanks! Should be G2G, here's a screenshot:

image

jairideout:
💯 Looks great!

PR #47, opened 2016-12-14
ENH: add plugin directory sphinx extension

jairideout:
This sphinx extension creates a directory listing plugins and their actions, similar to the QIIME 1 "script index" page.

Also tweaked the sidebar to display global TOC and searchbox instead of local TOC.

**Reviewers:** when merging, please choose the "rebase" option because these are two logically separate commits I'd like to retain.


source/theme/static/style.css
```diff
@@ -52,6 +52,10 @@ code.docutils.literal {
     color: #333;
 }
 
+.action-signature th[colspan] {
+    text-align: center;
+}
+
```
ebolyen: Add comments to group these

jairideout:
Thanks @ebolyen and @thermokarst for the reviews!

PR #59, opened 2017-01-06
ENH: use explicit .qza/.qzv extension in command outputs

### LINKED ISSUE #41: should commands use explicit .qza/.qzv extension for output files?
@thermokarst and I were chatting about this today. Commands in the docs are a little magical because output artifacts/visualizations aren't named with a `.qza`/`.qzv` extension. When only specifying the basename of the output file (sans extension), it isn't clear how the .qza/.qzv extension factors into output files. It also is hard to determine if the output filename is an artifact or visualization b/c the extension isn't in the output name. Using the basename only also makes it ambiguous whether the output is being saved as a directory or a file without an extension.

I had to explain this behavior several times during the Phoenix workshop, and it seemed to confuse users because .qza/.qzv files were a new concept that was just introduced, and the commands don't explicitly reflect how those extensions relate to output files.

Can we use explicit file extensions in the docs, and note somewhere that if the extension is omitted, it will automatically be appended?

jairideout:
Tested this locally with `make html`.

Fixes #41.

### DIFF
--- a/source/interfaces/artifact-api.rst
+++ b/source/interfaces/artifact-api.rst
@@ -110,6 +110,6 @@ Another powerful feature of QIIME 2 is that you can combine interfaces. For exam
 .. command-block::
    :no-exec:
 
-   qiime diversity alpha-group-significance --i-alpha-diversity oo.qza --m-metadata-file sample-metadata.tsv  --o-visualization oo-group-significance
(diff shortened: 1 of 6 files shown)


jairideout:
Thanks @ebolyen!!

PR #62, opened 2017-01-13
ENH: install latest dada2 via Bioconductor

jairideout:
Also updates conda install commands to specify ``--override-channels`` and ``-c defaults`` to avoid issues with channels in ``.condarc``.

See related discussion in https://github.com/qiime2/q2-dada2/issues/13


jairideout:
Thanks @gregcaporaso! Yep, all `conda install` commands in the docs now provide `--override-channels` along with `-c defaults`.

PR #71, opened 2017-01-27
ENH: "Choose your own adventure"-style downloads

### LINKED ISSUE #35: investigate why `make html` fails on certain machines
@thermokarst reported issues with `make html` on his macOS machine and on various Linux VMs. It sounds like commands marked with `:no-exec:` are being executed. @thermokarst is going to work with me to debug this issue. Note that @gregcaporaso and I didn't have this issue on our OS X machines.

### LINKED ISSUE #36: cURL commands for retrieval of MVP tutorial data are cURL version dependent
It appears that different versions of cURL handle redirects differently. `wget` works, but it does not ship with OSX by default. Versions of cURL observed to have the problem:

Linux:

```bash
$ curl --version
curl 7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.15.3 zlib/1.2.3 libidn/1.18 libssh2/1.4.2
Protocols: tftp ftp telnet dict ldap ldaps http file https ftps scp sftp 
Features: GSS-Negotiate IDN IPv6 Largefile NTLM SSL libz
``` 

OSX (Sierra):

```bash
curl --version
curl 7.49.1 (x86_64-apple-darwin16.0) libcurl/7.49.1 SecureTransport zlib/1.2.8
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtsp smb smbs smtp smtps telnet tftp 
Features: AsynchDNS IPv6 Largefile GSS-API Kerberos SPNEGO NTLM NTLM_WB SSL libz UnixSockets
```

### LINKED ISSUE #45: `make html` fails on macOS 10.12 due to conda-installed curl
The `r` packages installed as part of bootstrapping a QIIME2 environment apparently have a `curl` requirement. It looks like the recipe uses my system-installed curl, but, conda rewrites the certs path. Running `make html` with the `-s` removed on a curl command:

```sh
Extension error:
Command 'curl https://codeload.github.com/qiime2/q2studio/zip/0.0.6 -o q2studio-0.0.6.zip' exited with non-zero return code 77.

stdout:



stderr:

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  curl: (77) error setting certificate verify locations:
  CAfile: /Users/matthew/miniconda3/envs/q206/etc/pki/tls/certs/cacert.pem
  CApath: none
```

We have had a few `curl`-related issues pop up lately, maybe it is worth looking at `wget` instead?

### LINKED ISSUE #65: Add Sphinx directive for displaying download/fetch options
Related to #36

thermokarst:
Fixes #65 
Fixes #45 
Fixes #36 
Fixes #35

This adds support for download-method variations --- manual (browser),
`wget`, and `curl`. The aim here is for us to stop prescribing download
methods (this has lead to a lot of confusion in the past with different
versions of `curl`, as well as different `curl` configurations), and to
allow users to choose the best fit from a variety of options.

Secondarily, this circumvents some of those same `curl` problems, by
using `wget` when building the docs.

screen shot 2017-01-27 at 9 44 58 am
screen shot 2017-01-27 at 9 45 05 am


source/sphinx_extensions/command_block/extension.py
```diff
@@ -40,17 +63,34 @@ class CommandBlockDirective(docutils.parsers.rst.Directive):
     has_content = True
```
jairideout: Interesting, I thought this declaration would make `.. download::` fail, since it doesn't have any content. Any ideas? Maybe it's worth removing the line if it isn't doing anything.

source/sphinx_extensions/command_block/extension.py
```diff
@@ -40,17 +63,34 @@ class CommandBlockDirective(docutils.parsers.rst.Directive):
...
-            self._get_literal_block_node(self.content)
-        ]
+        command_mode = True if self.name == 'command-block' else False
+        opts = self.options
+        download_opts = [k in opts for k in ['url', 'rename']]
+
+        if command_mode:
+            self.assert_has_content()
+            if any(download_opts):
+                raise self.error('command-block does not support the '
```
jairideout: Can you raise `sphinx.errors.ExtensionError` instead? That's what's used for error raising in other parts of this extension, and it's the recommended way to raise an error within an extension (see the note at the start of [this section](http://www.sphinx-doc.org/en/1.5.2/extdev/appapi.html#producing-messages-logging) of the docs).

source/sphinx_extensions/command_block/extension.py
```diff
@@ -40,17 +63,34 @@ class CommandBlockDirective(docutils.parsers.rst.Directive):
...
+                                 'following options: `url`, `rename`.')
+            commands = functools.reduce(self._parse_multiline_commands,
+                                        self.content, [])
+            nodes = [self._get_literal_block_node(self.content)]
+        else:
+            if self.content:
+                raise self.error('Content block not supported for the '
```
jairideout: same comment here re: `ExtensionError`

source/sphinx_extensions/command_block/extension.py
```diff
@@ -40,17 +63,34 @@ class CommandBlockDirective(docutils.parsers.rst.Directive):
...
+                                 'download directive.')
+            if not all(download_opts):
+                raise self.error('Missing options for the download directive. '
```
jairideout: same comment here re: `ExtensionError`

source/sphinx_extensions/command_block/extension.py
```diff
@@ -40,17 +63,34 @@ class CommandBlockDirective(docutils.parsers.rst.Directive):
     has_content = True
 
     option_spec = {
-        'no-exec': docutils.parsers.rst.directives.flag
+        'no-exec': docutils.parsers.rst.directives.flag,
+        'url': docutils.parsers.rst.directives.unchanged_required,
+        'rename': docutils.parsers.rst.directives.unchanged_required,
```
jairideout: What do you think of using `saveas` instead of `rename`? It's a little clearer to me, but I also don't have strong feelings on this option's name.

source/sphinx_extensions/command_block/extension.py
thermokarst: Will look into this --- stay tuned!

source/sphinx_extensions/command_block/extension.py
thermokarst: 👍 

Background on `rename` choice: was originally going to implement a "rename" using the HTML `a` tag attr `download`, which can change the name of the remote file. It didn't pan out, due to Cross-Origin issues.

source/sphinx_extensions/command_block/extension.py
thermokarst: image

Oh I meant to chat with you about this! I don't think that `sphinx.errors.ExtensionError` is the right error to use here. The `CommandBlockDirective` is not sphinx-based, but rather, docutils-based. Cruising through the docutils [docs](http://docutils.sourceforge.net/docs/howto/rst-directives.html#error-handling), this looks like the right way to raise exceptions related to custom nodes/directives. Also, the trackback is a bit more informative:

```sh
(q2dev) ☁  docs [issue-65] ⚡ make preview
rm -rf build/*
sphinx-build -b dirhtml -D command_block_no_exec=1 -d build/doctrees -W source build/preview
Running Sphinx v1.4.6
making output directory...
loading pickled environment... not yet created
building [mo]: targets for 0 po files that are out of date
building [dirhtml]: targets for 27 source files that are out of date
updating environment: 85 added, 0 changed, 0 removed
reading sources... [100%] tutorials/moving-pictures

Warning, treated as error:
/Users/matthew/caporaso/qiime2/docs/source/interfaces/q2studio.rst:14: ERROR: Missing options for the download directive. Please specify `url` and `rename`.

make: *** [preview] Error 1
```

(note the file and line number leading you to the source of the error)

vs

```sh
(q2dev) ☁  docs [issue-65] ⚡ make preview
rm -rf build/*
sphinx-build -b dirhtml -D command_block_no_exec=1 -d build/doctrees -W source build/preview
Running Sphinx v1.4.6
making output directory...
loading pickled environment... not yet created
building [mo]: targets for 0 po files that are out of date
building [dirhtml]: targets for 27 source files that are out of date
updating environment: 85 added, 0 changed, 0 removed
reading sources... [ 16%] interfaces/q2studio
Extension error:
Missing options for the download directive. Please specify `url` and `rename`.
make: *** [preview] Error 1
```

The [docs](http://www.sphinx-doc.org/en/1.5.1/extdev/appapi.html#sphinx.errors.ExtensionError) for the `sphinx.errors.ExtensionError` indicate that it should be "Used for errors in setting up extensions," which from my perspective isn't really what is going on in the `run` method --- personally I think we should update the existing uses of `ExtensionError` to use the docutils-based exception, instead. If you agree, I can do that in this PR, or save it for a 🌧 day.

source/sphinx_extensions/command_block/extension.py
thermokarst: From the docutils docs:

```
    - `has_content`: A boolean; True if content is allowed.  Client
      code must handle the case where content is required but not
      supplied (an empty content list will be supplied).
```

So looks like for our purposes, this does nothing. Also, the `assert_has_content` uses that attr, which does us no good.

thermokarst: I spoke too soon. Apparently this *is* important. I suspect Sphinx might make use of it in ways I do not understand. Ultimately, my findings are: it stays! :clinking_glasses:

jairideout: Sounds good :P Thanks for investigating.

thermokarst:
Okay @jairideout, ready for another pass!

source/sphinx_extensions/command_block/templates/download.html
```diff
@@ -0,0 +1,53 @@
...
+
+    <div class="tab-content">
+      <div role="tabpanel" class="tab-pane fade in active" id="browser-{{ node.id }}">
+        <p>
+          <strong>Download URL:</strong> <a href="{{ node.url }}">
+            {{ node.url }}
+          </a>
+        </p>
+        <p>
+          <strong>Save as:</strong> {{ node.rename }}
```
jairideout: I think this should be `node.saveas` (it's used in two other places in this file as well).

jairideout:
Woooooo, thanks @thermokarst!

PR #79, opened 2017-02-02
DOC: adds importing of paired-end data

gregcaporaso:
The URLs to be downloaded don't exist yet, but I have the files locally and I've tested the import commands.

### DIFF
--- a/source/tutorials/importing.rst
+++ b/source/tutorials/importing.rst
@@ -14,28 +14,28 @@ Importing can be accomplished using any of the QIIME 2 :doc:`interfaces <../inte
 Sequence data
 -------------
 
-"EMP protocol" multiplexed fastq
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+"EMP protocol" multiplexed single-end fastq
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Format description
 ******************
 
-In the "Earth Microbiome Project (EMP) protocol" format, there are two ``fastq.gz`` files, one containing sequence reads and one containing the associated barcode reads, with the sequence data still multiplexed. The order of the records in the two ``fastq.gz`` files defines the association between a sequence read and its barcode read.
+In the "Earth Microbiome Project (EMP) protocol" format for single-end reads, there are two ``fastq.gz`` files, one containing sequence reads and one containing the associated barcode reads, with the sequence data still multiplexed. The order of the records in the two ``fastq.gz`` files defines the association between a sequence read and its barcode read.
 
 Obtaining example data
 **********************
 
 .. command-block::
 
-   mkdir raw-sequences
+   mkdir emp-single-end-sequences
 
 .. download::
    :url: https://data.qiime2.org/2.0.6/tutorials/moving-pictures/raw-sequences/barcodes.fastq.gz
-   :saveas: raw-sequences/barcodes.fastq.gz
+   :saveas: emp-single-end-sequences/barcodes.fastq.gz
 
 .. download::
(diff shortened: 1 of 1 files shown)


source/tutorials/importing.rst
```diff
@@ -44,8 +44,45 @@ Importing data
...
+ .. command-block::
+
+    mkdir emp-paired-end-sequences
+
+ .. download::
+    :url: https://data.qiime2.org/2017.02/tutorials/atacama/emp-paired-end-sequences-1p/forward.fastq.gz
+    :saveas: emp-paired-end-sequences/forward.fastq.gz
+
+.. download::
+    :url: https://data.qiime2.org/2017.02/tutorials/atacama/emp-paired-end-sequences-1p/reverse.fastq.gz
```
jairideout: These two directive options must be aligned with the "d" in "download". As-is, Sphinx is interpreting those options as directive content (this is causing the Travis failure).

source/tutorials/importing.rst
```diff
@@ -44,8 +44,45 @@ Importing data
...
+******************
+
+In the "Earth Microbiome Project (EMP) protocol" format for paired-end reads, there are three ``fastq.gz`` files, one containing forward sequence reads, one containing reverse sequence reads, and one containing the associated barcode reads, with the sequence data still multiplexed. The order of the records in the three ``fastq.gz`` files defines the association between the sequences reads and barcode reads.
+
+Obtaining example data
+**********************
+
+.. command-block::
+
+  mkdir emp-paired-end-sequences
```
jairideout: Align this with the "c" in "command-block" (this has caused rendering issues in the past). It looks like all `command-block` and `download` directives you added are misaligned -- the directive options and content must align with the start of the directive keyword.

source/tutorials/importing.rst
```diff
@@ -44,8 +44,45 @@ Importing data
...
+
+.. download::
+  :url: https://data.qiime2.org/2017.02/tutorials/atacama/emp-paired-end-sequences-1p/forward.fastq.gz
```
jairideout: `s/2017.02/2017.2/g`

PEP 440 requires us to drop the leading zero.

source/tutorials/importing.rst
```diff
@@ -76,6 +113,36 @@ Importing data
...
+Format description
+******************
+
+In this format, there are two ``fastq.gz`` file for each sample in the study, and the file name includes the sample identifier. The forward and reverse read file names for a single sample might look like ``L2S357_15_L001_R1_001.fastq.gz`` and ``L2S357_15_L001_R2_001.fastq.gz``, respectively. The underscore-separated fields in this file name are the sample identifier, the barcode sequence or a barcode identifier, the lane number, the read number, and the set number.
+
+Obtaining example data
+**********************
+
+.. download::
+  :url: https://data.qiime2.org/2017.02/tutorials/importing-sequence-data/casava-18-paired-end-demultiplexed.zip
```
jairideout: The tutorial is now called `importing` instead of `importing-sequence-data`, can you update the URL? The single-end URL above still references the 2.0.6 tutorial name but it'll get updated for the release.

source/tutorials/importing.rst
```diff
@@ -76,6 +113,36 @@ Importing data
...
+
+Importing data
+**************
+
+.. command-block::
+  qiime tools import \
+    --type 'SampleData[PairedEndSequencesWithQuality]' \
+    --input-path casava-18-paired-end-demultiplexed \
+    --source-format CasavaOneEightSingleLanePerSampleDirFmt \
+    --output-path demux.qza
```
jairideout: This command will fail when run with `make html` because the output name has already been used (overwriting outputs is disallowed by `command-block`).

jairideout:
🌚

PR #83, opened 2017-02-03
ENH: replaces 88 soils tutorial with Atacama soils tutorial

gregcaporaso:
**DO NOT MERGE**

This tutorial still needs some work, but the commands here will let us do integration tests. 

I'd also like to update the sample metadata to make it a little cleaner, but that can happen after this tutorial goes in.


jairideout:
Merging now before @thermokarst and I rework the data.qiime2.org links. Our goal is to get `make html` to pass so we can do integration testing. We can add more text, etc in future PRs.

PR #85, opened 2017-02-06
ENH: Dynamic Q2 URL Versioning

thermokarst:
Not quite there --- please do not merge!


jairideout:
Chatted with @thermokarst. We're closing this in favor of keeping manual URL version upgrades (e.g. search and replace) so we can manage dev vs release versions more easily.

PR #93, opened 2017-02-08
ENH: QIIME 2 Upgrade/Update Instructions

thermokarst:


source/upgrade.rst
```diff
@@ -0,0 +1,74 @@
...
+===================================
+
+Existing QIIME 2 Core distribution installations can be either upgraded (i.e. installing a newer release of QIIME 2) or updated (i.e. patch updates within a given release of QIIME 2). Please see the section below that most closely matches your existing QIIME 2 installation.
+
+.. _upgrading:
+
+Upgrading QIIME 2 Core 2017.2
+-----------------------------
+
+Native Installation
```
gregcaporaso: Should these be sub-sub-headings (underlined with `~`)?

source/upgrade.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Amazon AWS
+    Please see the :doc:`Amazon AWS install doc <install/virtual/aws>` for instructions on how to set up a new QIIME 2 Core 2017.2 AWS instance.
+
+.. _updating:
+
+Updating QIIME 2 Core 2017.2
+----------------------------
+
+macOS/OS X (Native Installation)
```
gregcaporaso: Same comment as above about sub-sub-headings.

source/upgrade.rst
```diff
@@ -0,0 +1,74 @@
...
+    In a terminal run the following commands:
+
+    .. command-block::
+       :no-exec:
+
+       conda update conda
+       # If you named your conda environment something else, use that name here.
```
gregcaporaso: *something else when you initially installed QIIME 2, ...*

gregcaporaso: same comment in the other locations below...

source/upgrade.rst
thermokarst: Clarified offline: these are definition lists, they look good when the docs are built

source/upgrade.rst
thermokarst: Same as above.

Issue #120, opened 2017-03-09
display plugin short/long descriptions in plugin directory Sphinx extension


jairideout:
The QIIME 2 framework and q2cli now support plugin short/long descriptions (thanks @maxvonhippel for hooking that up!). It would be great to include these descriptions in the [plugin directory Sphinx extension](https://github.com/qiime2/docs/tree/master/source/sphinx_extensions/plugin_directory) that auto-generates plugin docs.

There's an example of the rendered plugin directory docs [here](https://docs.qiime2.org/2017.2/plugins/available/). We'd want the "long" description at the top of the page, in the same group as version, citation text, etc. (e.g. check out [this page](https://docs.qiime2.org/2017.2/plugins/available/diversity/)). We'd want the short description to appear in the title heading of that page (e.g. "diversity: short_description", or just "diversity" if there isn't a short description).

@maxvonhippel, let me know if you need clarification on where to display these descriptions (I can post screenshots if it helps).

maxvonhippel:
Sounds great!  I'll look through this sometime tomorrow evening or next week and let you know if I have any questions!

PR #121, opened 2017-03-11
Plugin descriptions

maxvonhippel:
Added automatic sphinx handling of `description` and `short_description` per issue [#120](https://github.com/qiime2/docs/issues/120#issuecomment-285428117).

See below screenshots:

**Plugins page**
Note that if no `short_description` exists for the plugin then I just show the name of the plugin.  If a short description exists, I append `: short_description` to the title of the item in the `.doctree`.
plugins_page

**Available Plugins page**
Note that the approach taken on the plugins page as described above automatically propagates to this page.
available_plugins

**feature-table page**
Note the implementation of the `description` parameter in the table of values, per the suggestion in the referenced issue.  Let me know if you'd like this tweaked in any particular way.
feature-table

**types page**
Currently if no `short_description` exist there is a hard-coded default value which includes the URL.  This is great for the command-line output and is *by design* per our previous conversations on [qiime2](https://github.com/qiime2/qiime2/pull/222), [q2cli](https://github.com/qiime2/q2cli/pull/130#issuecomment-285425486), and [q2-feature-table](https://github.com/qiime2/q2-feature-table/pull/91#event-988543716).  However, it is a bit redundant in this context.  What do you think?  If you'd prefer to just show "No description available" or something, I can fairly easily implement a REGEX to detect this scenario and respond accordingly.
types

### DIFF
--- a/source/sphinx_extensions/plugin_directory/templates/available.rst
+++ b/source/sphinx_extensions/plugin_directory/templates/available.rst
@@ -6,9 +6,9 @@ QIIME 2 microbiome analysis functionality is made available to users via plugins
 .. toctree::
    :maxdepth: 3
 
-   {% for name, _ in plugins|dictsort %}
-   {% if _.short_description %}
-   {{ name }}: {{ _.short_description }} <{{ name }}/index>
+   {% for name, plugin in plugins|dictsort %}
+   {% if plugin.short_description %}
+   {{ name }}: {{ plugin.short_description }} <{{ name }}/index>
    {% else %}
    {{ name }}/index
    {% endif %}
--- a/source/sphinx_extensions/plugin_directory/templates/available.rst
+++ b/source/sphinx_extensions/plugin_directory/templates/available.rst
@@ -7,5 +7,6 @@ QIIME 2 microbiome analysis functionality is made available to users via plugins
    :maxdepth: 3
 
    {% for name, _ in plugins|dictsort %}
-   {{ name }}/index
+   {{ name }}: {{ _.short_description }} <{{ name }}/index>
    {% endfor %}
+
--- a/source/sphinx_extensions/plugin_directory/templates/available.rst
+++ b/source/sphinx_extensions/plugin_directory/templates/available.rst
@@ -7,6 +7,10 @@ QIIME 2 microbiome analysis functionality is made available to users via plugins
    :maxdepth: 3
 
    {% for name, _ in plugins|dictsort %}
+   {% if _.short_description %}
    {{ name }}: {{ _.short_description }} <{{ name }}/index>
+   {% else %}
+   {{ name }}/index
+   {% endif %}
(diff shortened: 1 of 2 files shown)


source/sphinx_extensions/plugin_directory/templates/available.rst
```diff
@@ -7,5 +7,10 @@ QIIME 2 microbiome analysis functionality is made available to users via plugins
    :maxdepth: 3
 
    {% for name, _ in plugins|dictsort %}
+   {% if _.short_description %}
```
jairideout: This is minor: can you change the variable name `_` to something more descriptive such as `plugin`? The original code didn't use the values of the dictionary while iterating, so we assigned to a throwaway variable named `_` (this is just convention in Python and doesn't have any special meaning). Now that the values of the dict are being used it's a good idea to give the variable a name.

source/sphinx_extensions/plugin_directory/templates/plugin.rst
```diff
@@ -6,6 +6,10 @@
    <table class="table plugin-info">
      <tbody>
        <tr>
+         <th scope="row">Description</th>
+         <td>{{ plugin.description|urlize }}</td>
```
jairideout: Can you take the same approach as the `Support` and `Citation` sections below? Since these are free text blocks, the text has to be split into lines, where each line is `urlized` individually (I tried it without breaking lines and `urlize` doesn't work as expected).

maxvonhippel: Done.  Yeah I also noticed that `urlize` can be strange sometimes.  Looks to me like it's working now that I changed the code to reflect the structure used in `Support` and `Citation` per your suggestion.

source/sphinx_extensions/plugin_directory/templates/available.rst
maxvonhippel: Good thinking.  I think `name`, `plugin` is a good set of things to call the key-value pair, because then calling the parameters with `plugin.param_name` reflects the way you'd do it if you had the actual plugin object in Python so it's a pretty natural way to think about it, and it's clear that the `name` is the key, so that naming convention seems both lucid and nicely pythonic.  Anyway, thanks for pointing that out, and I've updated the code accordingly.

maxvonhippel:
Awesome, just updated the code accordingly.  Tested on my laptop and appears to be working.  Let me know if you'd like anything else changed!  I agree that we can finesse the overall behavior later on.  I'd like to dial back a bit on reliance on the `/index` feature of the `.doctree` so I could get a bit more customized behavior on some of the different pages and avoid redundancy more.  This could be accomplished with custom titles, similar to how I differentiated printing for plugins with vs without short_descriptions.  Anyway, we can talk about this all later.  Thanks!

PR #123, opened 2017-03-15
ENH: fixes #118

### LINKED ISSUE #118: add note on how to import FeatureData[Sequence] and other data necessary for moving data from QIIME 1
This [came up on the forum](https://forum.qiime2.org/t/importing-fasta-qual-instead-of-fastq/215/3) and in discussion with one of my students this semester. This will be important for anyone who is running "upstream" processing steps in QIIME 1 (e.g., if they're working with 454 data) as the best point of entry into QIIME 2 for those users will be importing `FeatureTable[Frequency]` and `FeatureData[Sequence]` (their representative sequences). 

```bash
qiime tools import --input-path rep_set.fna --output-path temp/rep-seqs.qza --type FeatureData[Sequence]
```

gregcaporaso:
Addresses #118. 

The 2017.2 data urls for this tutorial are `2017.2/tutorials/importing-sequence-data/...`, which corresponds to the old name for this tutorial. The files that I'm adding here are in a new `2017.3/tutorials/importing` directory, which I think corresponds to our URL scheme for these data. Is that correct (in which case `2017.2/tutorials/import-sequence-data` would become `2017.3/tutorials/importing`)? 

This PR depends on https://github.com/qiime2/data.qiime2.org/pull/9

### DIFF
--- a/source/tutorials/importing.rst
+++ b/source/tutorials/importing.rst
@@ -19,8 +19,8 @@ You may want to begin by creating a directory to work in.
    mkdir qiime2-importing-tutorial
    cd qiime2-importing-tutorial
 
-Sequence data
--------------
+Sequence data with sequence quality information
+-----------------------------------------------
 
 "EMP protocol" multiplexed single-end fastq
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


thermokarst:
PR https://github.com/qiime2/data.qiime2.org/pull/9 has been merged.

gregcaporaso:
`make html` completed successfully locally.

PR #138, opened 2017-05-15
MAINT: Update VBox GA instructions

thermokarst:
Feel free to suggest alternative verbiage. This PR is necessitated by the changes proposed in https://github.com/qiime2/playbooks/pull/29, which mitigates the missing optical drive in VirtualBox by pre-installing the guest additions at image creation.

### DIFF
--- a/source/install/virtual/virtualbox.rst
+++ b/source/install/virtual/virtualbox.rst
@@ -74,7 +74,7 @@ Optional: Shared Folders and Copy-and-Paste
 1. Install Guest Additions
 ..........................
 
-Guest Additions are already installed in the virtual machine, but if you need to reinstall them, you can perform the following steps.
+**Guest Additions are already installed in the virtual machine, but if you need to reinstall them, you can perform the following steps.**
 
 After clicking on the CD icon in the lower left, click on ``Run Software`` in the upper right.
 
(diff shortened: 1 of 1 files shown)


ebolyen:
What if we made the title have something like `(Optional)` and then made your new sentence bold?

thermokarst:
Thanks @ebolyen, makes sense. The optional is already indicated in the [section header](https://github.com/qiime2/docs/pull/138/files#diff-d8f972fac892ddf520b94a93858c91f6R71), but I did make the new sentence bold. Do you think we should re-arrange where/how we indicate optional steps?

jairideout:
> What if we made the title have something like (Optional) and then made your new sentence bold?

What if instead of bolding the sentence, you made it a `.. note:` directive?

jairideout:
> Do you think we should re-arrange where/how we indicate optional steps?

I think the general flow of the section makes sense (I read through the live rendered page on docs.qiime2.org for context).

ebolyen:
I agree with the flow, I think you could still do the optional steps 2-N without needing to do step 1 though correct? That's why I was thinking this header would be "extra" optional. Maybe there's a better word?

thermokarst:
Rendered:
screen shot 2017-05-16 at 3 42 09 pm

I didn't put "Optional" in the `h3` but did add to the `info` box, because the whole section is optional, but this step might need to happen if someone needs to update/upgrade their guest additions.

PR #166, opened 2017-06-27
ENH: Mention that fastq supported in fastqmanifest

### LINKED ISSUE #162: Mention that `fastq` files get `gzip`d on import for `FastqManifestFormat`(s)
The importing doc (and possibly one or more other docs) say that users must `gzip` their `fastq` files before importing with any of the various `FastqManifestFormat`s.  This is not true; our code will `gzip` it for them.

maxvonhippel:
With this PR, the importing tutorial now clarifies that both `fastq` and `fasts.gz` filetypes are supported for importing from the various `fastqManifestFormat`s.  Fixes #162

### DIFF
--- a/source/tutorials/importing.rst
+++ b/source/tutorials/importing.rst
@@ -98,7 +98,7 @@ Importing data
 Format description
 ******************
 
-In the fastq manifest formats, a manifest file maps sample identifiers to ``fastq.gz`` `absolute filepaths`_ that contain sequence and quality data for the sample, and indicates the direction of the reads in each ``fastq.gz`` absolute filepath. The manifest file will generally be created by you, and it is designed to be a simple format that doesn't put restrictions on the naming of the demultiplexed ``fastq.gz`` files, since there is no broadly used naming convention for these files. There are no restrictions on the name of the manifest file.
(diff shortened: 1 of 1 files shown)


maxvonhippel:
Don't merge this yet, I messed up the zip specification.  Fixing currently.

maxvonhippel:
@ebolyen I agree re `/`, good idea.  As far as the compression is concerned, I am not sure where I stand.  On the one hand, of course mentioning what happens does to uncompressed `fastq` files upon import would help users understand .. that.  But on the other hand, exposing that functionality could lead to confusion.  What if users misunderstand and think that the files they input will get gzipped (as opposed to the copies of those files made by QIIME and stuck inside the artifact)?  It just seems a bit precarious to me and I am not sure how much explanation is too much.

ebolyen:
Let's not sweat the gzip explanation then. Re: your edit; the `/` should show up outside of the code/literal blocks, so I don't think it will be a problem.

maxvonhippel:
Sounds good.  I kept the `or` in the first paragraph (per our verbal discussion), and that way the subsequent `/`s should be very clear in meaning.

That said, if this still looks awkward to you, let me know and I will brainstorm alternatives.  Thanks!

PR #171, opened 2017-07-05
UX BUG: Native install (mac vs linux)

### LINKED ISSUE #169: UX BUG: Native install (mac vs linux)
This has come up several times recently: users are accidentally copying and pasting the mac os install commands when installing on a linux host, likely because the mac os commands occur first. Maybe we should replace this section with a tabbed panel (like the `download` blocks). The three tabs would be a default "instructions" page, instructing users to click either the mac tab or the linux tab, and then the other two tabs would be the aforementioned mac and linux tabs. This might help prevent the copy-paste issues, although maybe it poses some new problem that I am not considering. Thoughts?

maxvonhippel:
Fixes #169

### DIFF
--- a/source/install/native.rst
+++ b/source/install/native.rst
@@ -25,7 +25,7 @@ Once you have Miniconda installed, create a ``conda`` environment and install th
 
 .. raw:: html
    
-   <div class="tabbed panel panel-default">
+   <div class="tabbed">
       <ul class="nav nav-tabs">
          <li class="active"><a data-toggle="tab" href="#instructions">Instructions</a></li>
          <li><a data-toggle="tab" href="#macOS">macOS/OS X (64-bit)</a></li>
--- a/source/install/native.rst
+++ b/source/install/native.rst
@@ -24,34 +24,23 @@ Once you have Miniconda installed, create a ``conda`` environment and install th
 
 
 .. raw:: html
-
-   <style>
-      .tab-content {
-         border-left: 1px solid #ddd;
-         border-right: 1px solid #ddd;
-         border-bottom: 1px solid #ddd;
-         padding: 10px;
-      }
-      .nav-tabs {
-         margin-bottom: 0;
-      }
-   </style>
-   <ul class="nav nav-tabs">
-      <li class="active"><a data-toggle="tab" href="#instructions">Instructions</a></li>
-      <li><a data-toggle="tab" href="#macOS">macOS/OS X (64-bit)</a></li>
-      <li><a data-toggle="tab" href="#linux">Linux (64-bit)</a></li>
-   </ul>
-   <div class="tab-content">
-      <div id="instructions" class="tab-pane fade in active">
-         <p>The installation process differs slightly across platforms. From the above tabs, please choose the installation instructions that are appropriate for your platform.</p>
-      </div>
-      <div id="macOS" class="tab-pane fade">
-         <p>In a terminal run the following command:</p>
-         <pre><code>conda create -n qiime2-2017.6 --file https://data.qiime2.org/distro/core/qiime2-2017.6-conda-osx-64.txt</code></pre>
-      </div>
-      <div id="linux" class="tab-pane fade">
-         <p>In a terminal run the following command:</p>
-         <pre><code>conda create -n qiime2-2017.6 --file https://data.qiime2.org/distro/core/qiime2-2017.6-conda-linux-64.txt</code></pre>
+   
+   <div class="tabbed panel panel-default">
+      <ul class="nav nav-tabs">
+         <li class="active"><a data-toggle="tab" href="#instructions">Instructions</a></li>
+         <li><a data-toggle="tab" href="#macOS">macOS/OS X (64-bit)</a></li>
+         <li><a data-toggle="tab" href="#linux">Linux (64-bit)</a></li>
+      </ul>
+      <div class="tab-content">
+         <div id="instructions" class="tab-pane fade in active">
+            <p>From the above tabs, please choose the installation instructions that are appropriate for your platform.</p>
+         </div>
+         <div id="macOS" class="tab-pane fade">
+            <pre>conda create -n qiime2-2017.6 --file https://data.qiime2.org/distro/core/qiime2-2017.6-conda-osx-64.txt</pre>
+         </div>
+         <div id="linux" class="tab-pane fade">
+            <pre>conda create -n qiime2-2017.6 --file https://data.qiime2.org/distro/core/qiime2-2017.6-conda-linux-64.txt</pre>
+         </div>
       </div>
    </div>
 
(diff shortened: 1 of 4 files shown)


maxvonhippel:
This is ready for review.

maxvonhippel:
Note that, due to the `hint`, I thought it made sense to just use the two tabs (`macOS` and `linux`), and not include a third `default` one as @thermokarst suggested.  That said, if the consensus is that a `default` tab is the way to go, I'm happy to implement that.  So to whoever reviews this - could you please look at the current UX, and then read the original proposal [here](https://github.com/qiime2/docs/issues/169), and let me know what you think?

maxvonhippel:
Additionally, should this style be implemented in the `upgrade` page as well?  If so, I think it would make sense for me to do so in this same PR.  Thanks for the feedback!

maxvonhippel:
This is ready for review.

source/install/native.rst
```diff
@@ -22,23 +22,39 @@ Install QIIME 2 within a ``conda`` environment
...
+      <li><a data-toggle="tab" href="#macOS">macOS/OS X (64-bit)</a></li>
+      <li><a data-toggle="tab" href="#linux">Linux (64-bit)</a></li>
+   </ul>
+   <div class="tab-content">
+      <div id="instructions" class="tab-pane fade in active">
+         <p>The installation process differs slightly across platforms. From the above tabs, please choose the installation instructions that are appropriate for your platform.</p>
+      </div>
+      <div id="macOS" class="tab-pane fade">
+         <p>In a terminal run the following command:</p>
+         <pre><code>conda create -n qiime2-2017.6 --file https://data.qiime2.org/distro/core/qiime2-2017.6-conda-osx-64.txt</code></pre>
```
thermokarst: Can you remove the `<code>` tags? It prevents the `pre` container from scrolling, which forces line breaks into the command:

screen shot 2017-07-06 at 6 03 43 am

vs

screen shot 2017-07-06 at 6 03 08 am

source/install/native.rst
```diff
@@ -22,23 +22,39 @@ Install QIIME 2 within a ``conda`` environment
...
+      <li class="active"><a data-toggle="tab" href="#instructions">Instructions</a></li>
+      <li><a data-toggle="tab" href="#macOS">macOS/OS X (64-bit)</a></li>
+      <li><a data-toggle="tab" href="#linux">Linux (64-bit)</a></li>
+   </ul>
+   <div class="tab-content">
+      <div id="instructions" class="tab-pane fade in active">
+         <p>The installation process differs slightly across platforms. From the above tabs, please choose the installation instructions that are appropriate for your platform.</p>
+      </div>
+      <div id="macOS" class="tab-pane fade">
+         <p>In a terminal run the following command:</p>
```
thermokarst: This feels unnecessary to me (the tutorial starts out with prompting users to open a terminal). If you think it makes sense to include, is there some way to pull it out of the tabbed pane and into the paragraph above?

source/install/native.rst
```diff
@@ -22,23 +22,39 @@ Install QIIME 2 within a ``conda`` environment
 
 Once you have Miniconda installed, create a ``conda`` environment and install the QIIME 2 Core 2017.6 distribution within the environment. We **highly** recommend creating a *new* environment specifically for the QIIME 2 release being installed, as there are many required dependencies that you may not want added to an existing environment. You can choose whatever name you'd like for the environment. In this example, we'll name the environment ``qiime2-2017.6`` to indicate what QIIME 2 release is installed (i.e. ``2017.6``).
 
-The installation process differs slightly across platforms. Please choose the installation instructions that are appropriate for your platform.
 
-macOS/OS X (64-bit)
-~~~~~~~~~~~~~~~~~~~
+.. raw:: html
+
+   <style>
```
thermokarst: Can you move this CSS into the [stylesheet](https://github.com/qiime2/docs/blob/master/source/theme/static/style.css)? Perhaps you can refactor the [`download` directives](https://github.com/qiime2/docs/blob/master/source/theme/static/style.css#L92-L99) to share the same styles? Also, there is some extra padding at the bottom, can you make sure that gets removed (hint: reusing the existing styles should help with that)?

source/install/native.rst
```diff
@@ -22,23 +22,39 @@ Install QIIME 2 within a ``conda`` environment
...
+      }
+   </style>
+   <ul class="nav nav-tabs">
+      <li class="active"><a data-toggle="tab" href="#instructions">Instructions</a></li>
+      <li><a data-toggle="tab" href="#macOS">macOS/OS X (64-bit)</a></li>
+      <li><a data-toggle="tab" href="#linux">Linux (64-bit)</a></li>
+   </ul>
+   <div class="tab-content">
+      <div id="instructions" class="tab-pane fade in active">
+         <p>The installation process differs slightly across platforms. From the above tabs, please choose the installation instructions that are appropriate for your platform.</p>
```
thermokarst: Nice! How about we drop the first sentence?

>  From the above tabs, please choose the installation instructions that are appropriate for your platform.

source/install/native.rst
thermokarst: Sorry, this doesn't directly answer your question from [above](https://github.com/qiime2/docs/pull/171#issuecomment-313246454), but let me know what you think.

source/install/native.rst
maxvonhippel: Good idea!

source/install/native.rst
maxvonhippel: I think you're right.

maxvonhippel:
Ok, this is ready for review again @thermokarst .  Thanks!

source/theme/static/style.css
```diff
@@ -79,17 +79,16 @@ div.admonition.qiime1 {
     text-align: center;
 }
 
-/* `download` extension */
-.download .white-bg {
+/* `tabbed` extension */
```
thermokarst: This comment doesn't really make sense anymore --- maybe `tabbed content`?

maxvonhippel: Good catch, thanks!  ... and done!

maxvonhippel:
Ok, this is ready for a final pass

source/install/native.rst
```diff
@@ -22,23 +22,28 @@ Install QIIME 2 within a ``conda`` environment
 
 Once you have Miniconda installed, create a ``conda`` environment and install the QIIME 2 Core 2017.6 distribution within the environment. We **highly** recommend creating a *new* environment specifically for the QIIME 2 release being installed, as there are many required dependencies that you may not want added to an existing environment. You can choose whatever name you'd like for the environment. In this example, we'll name the environment ``qiime2-2017.6`` to indicate what QIIME 2 release is installed (i.e. ``2017.6``).
 
-The installation process differs slightly across platforms. Please choose the installation instructions that are appropriate for your platform.
 
-macOS/OS X (64-bit)
-~~~~~~~~~~~~~~~~~~~
+.. raw:: html
+   
+   <div class="tabbed panel panel-default">
```
thermokarst: `<div class="tabbed panel panel-default">` -> `<div class="tabbed">`

Semantically this isn't a Bootstrap `panel`.

maxvonhippel: In that case, would 

`<div class="tabbed">`

fit the bill?

maxvonhippel: Good catch!  And done.  Thanks!

thermokarst:
Looks good! Thanks @maxvonhippel!! 🌮 🌮

PR #195, opened 2017-08-28
MAINT: update to 2017.8

jairideout:
Updated:

- references to `2017.7` -> `2017.8`
- removed forum data links in favor of data.qiime2.org links
- updated scikit-learn version `0.18.2` -> `0.19.0`
- verified even sampling depth in Moving Pictures tutorial matches what is in
  the text
- added `q2-sample-classifier` and `q2-longitudinal` to Core distribution list


jairideout:
Adding "do not merge" until the new classifiers are trained and uploaded (we need their MD5 sums in the docs). Good catch @ebolyen!

I'm training the classifiers right now and they are almost finished :tada:

PR #183, opened 2017-07-31
q2-sample-classifier tutorial

nbokulich:

PR #197, opened 2017-08-29
MAINT: sample-classifier tutorial style-match

thermokarst:


thermokarst:
Missed this in the review of https://github.com/qiime2/docs/pull/183.

PR #204, opened 2017-09-22
added tutorial for longitudinal nmit

nbokulich:
DO NOT MERGE: pending [this PR](https://github.com/qiime2/q2-longitudinal/pull/55)

### DIFF
--- a/source/tutorials/longitudinal.rst
+++ b/source/tutorials/longitudinal.rst
@@ -135,21 +135,21 @@ Now we are ready run NMIT. The output of this command is a distance matrix that 
 
    qiime longitudinal nmit \
      --i-table ecam-table-taxa.qza \
-     --m-metadata-file ecam_map_maturity.txt \
+     --m-metadata-file ecam-sample-metadata.tsv \
      --p-individual-id-column studyid \
-     --o-distance-matrix nmit-dm.qza \
-     --p-corr-method pearson
+     --p-corr-method pearson \
+     --o-distance-matrix nmit-dm.qza
 
 
 Now let's put that distance matrix to work. First we will perform PERMANOVA tests to evaluate whether between-group distances are larger than within-group distance.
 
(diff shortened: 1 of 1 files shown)


source/tutorials/longitudinal.rst
```diff
@@ -114,5 +114,61 @@ Second, control charts display the mean value of "metric" at each "state". The f
...
+   qiime emperor plot \
+     --i-pcoa nmit-pc.qza \
+     --m-metadata-file ecam_map_maturity.txt \
+     --o-visualization nmit-emperor.qzv
+
+So there it is. We can use PERMANOVA test or other distance-based statistical tests to determine whether groups exhibit different longitudinal microbial interdependence relationships, and PCoA/emperor to visualize the relationships among groups of subjects. **Don't forget the caveats mentioned above about using and interpreting NMIT**. Now be safe and have fun.
+
 .. _ECAM study: https://doi.org/10.1126/scitranslmed.aad7121
 .. _statsmodels LME description page: http://www.statsmodels.org/dev/mixed_linear.html
+.. _Zhang et al., 2015: https://doi.org/10.1002/gepi.22065
```
elong0527: The reference year should be "2017"

source/tutorials/longitudinal.rst
```diff
@@ -114,5 +114,61 @@ Second, control charts display the mean value of "metric" at each "state". The f
...
+
+.. note:: NMIT, as with most longitudinal methods, largely depends on the quality of the input data. This method will only work for longitudinal data (i.e., the same subjects are sampled repeatedly over time). To make the method robust, we suggest a minimum of 5-6 samples (time points) per subject, but the more the merrier. NMIT does not require that samples are collected at identical time points (and hence is robust to missing samples) but this may impact data quality if highly undersampled subjects are included, or if subjects' sampling times do not overlap in biologically meaningful ways. It is up to the users to ensure that their data are high quality and the methods are used in a biologically relevant fashion.
+
+.. note:: NMIT can take a long time to run; the following command takes around an hour to run. Just download the results file displayed below and skip ahead if you want to see how NMIT data are used downstream in QIIME2.
```
elong0527: We should show the example in genus level and screen out OTUs who have 0s in most of counts. 

In the paper, we only considered OTUs who have relative abundance >0.1% in more than 15% of the total samples. That result in 23 major taxa out of 498 taxa in genus level.

nbokulich: Thanks @elong0527 ! Do you want to send me the files that you used in the paper? We could just upload these very same files into this tutorial.

source/tutorials/longitudinal.rst
nbokulich: 👍 thanks for catching this @elong0527 ! I have fixed this now.

source/tutorials/longitudinal.rst
elong0527: I think we can use the current tutorial data instead of maintaining two sets of data. 

That will be great if you could add one more step to make the taxa table in Genus level. If it is slow to run (true in R), we can just use the current Species level (I guess) taxa table to illustrate NMIT. 

The screening step has been implemented in the code now. 

[nbokulich/q2-intervention/pull/3#issuecomment-331730846](https://github.com/nbokulich/q2-intervention/pull/3#issuecomment-331730846)

nbokulich: Hi @elong0527 the tutorial data that we use currently (species-level data) are only used for the NMIT example. So I do not have any problem with replacing this file with a genus-level taxa feature table (including the additional filtering that you describe). Thanks!

nbokulich:
Thanks @thermokarst ! Looks great.

PR #206, opened 2017-09-27
ENH: plugin directory supports multiple interfaces

jairideout:
The plugin directory now displays CLI and Artifact API help text for actions instead of a custom table of inputs, parameters, and outputs (this table was generally confusing to users).

Thanks @thermokarst for having the idea of using the help text that's already being generated for the CLI and Artifact API! The code is much simpler now too :)

Also, instead of using an action's underscored name (`Action.id`), the action's CLI name (dashed) is used in URLs and content.

# Examples

## `make preview`/`make html` output

Indicates progress since it can take a little while to generate the plugin directory:

image

## CLI help text (default view)

image

## Artifact API help text

image

## New URL scheme for actions

**Before:** https://docs.qiime2.org/2017.8/plugins/available/demux/emp_paired/

**After:** https://docs.qiime2.org/2017.9/plugins/available/demux/emp-paired/

:shipit:

### DIFF
--- a/source/sphinx_extensions/plugin_directory/extension.py
+++ b/source/sphinx_extensions/plugin_directory/extension.py
@@ -9,12 +9,22 @@
 import os
 import os.path
 import shutil
+import subprocess
+import textwrap
 
 import jinja2
 import qiime2.sdk
 
 
 def generate_rst(app):
+    app.info("Generating QIIME 2 plugin directory... (this may take awhile)")
+
+    # Refresh the CLI cache just in case it is out of date with what's
+    # installed (this should only affect packages while they are installed in
+    # development mode). The CLI is used below to generate some of the help
+    # text.
+    subprocess.run(['qiime', 'dev', 'refresh-cache'], check=True)
+
     plugins = qiime2.sdk.PluginManager().plugins
     loader = jinja2.PackageLoader('sphinx_extensions.plugin_directory',
                                   'templates')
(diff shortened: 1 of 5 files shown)


source/sphinx_extensions/plugin_directory/extension.py
```diff
@@ -32,47 +42,41 @@ def generate_rst(app):
         fh.write(rendered)
 
     for plugin in plugins.values():
-        plugin_dir = os.path.join(rst_dir, plugin.name)
+        plugin_cli_name = plugin.name.replace('_', '-')
```
jairideout: I updated the code and templates to use CLI-style names (i.e. dashed names) for plugins. Plugins are typically named with dashes already (e.g. `'feature-table'` for `Plugin.name`) but that is by convention only, so there could be non-dashed names in the wild. This change is mainly to future-proof the code.

Action names are also converted to CLI-style names (see below).

source/sphinx_extensions/plugin_directory/templates/plugin.rst
```diff
@@ -8,9 +8,9 @@
        <tr>
          <th scope="row">Description</th>
          <td>
-         	{% for line in plugin.description.splitlines() %}
-         	{{ line|urlize }}<br/>
-         	{% endfor %}
+           {% for line in plugin.description.splitlines() %}
```
jairideout: There were some tabs in this file, replaced with spaces.

source/theme/static/style.css
```diff
@@ -74,11 +74,6 @@ div.admonition.qiime1 {
     border: 1px solid #FAEBCC;
 }
 
-/* `plugin_directory` extension */
```
jairideout: No longer needed because HTML table is gone!

source/sphinx_extensions/plugin_directory/templates/plugin.rst
thermokarst: 📑

source/sphinx_extensions/plugin_directory/templates/action.rst
```diff
@@ -1,48 +1,21 @@
...
-   </table>
+   <div class="tabbed">
+      <ul class="nav nav-tabs">
+         <li class="active"><a data-toggle="tab" href="#cli">Command line interface</a></li>
+         <li><a data-toggle="tab" href="#api">Artifact API</a></li>
+      </ul>
+      <div class="tab-content">
+         <div id="cli" class="tab-pane fade in active">
+            <pre>
+{{ cli_help }}   </pre>
```
jairideout: The trailing `</pre>` tags are gross -- I couldn't find a way to have each line in `cli_help` indented by at least 3 spaces so that it's valid RST, while also not adding trailing blank lines in the rendered `<pre>` content (which is what happens if I put `</pre>` on its own line). If anyone knows how to clean this up that would be awesome!

thermokarst: How about using the jinja whitespace control flags (`-`) to strip leading and trailing whitespace, plus a judicious `strip()` for good measure:

```diff
diff --git a/source/sphinx_extensions/plugin_directory/extension.py b/source/sphinx_extensions/plugin_directory/extension.py
index 11ff4d0..1eafc46 100644
--- a/source/sphinx_extensions/plugin_directory/extension.py
+++ b/source/sphinx_extensions/plugin_directory/extension.py
@@ -67,7 +67,7 @@ def generate_rst(app):
                                       stdout=subprocess.PIPE,
                                       stderr=subprocess.STDOUT)
                 cli_help = textwrap.indent(proc.stdout.decode('utf-8'),
-                                           directive_indent)
+                                           directive_indent).strip()
                 api_help = textwrap.indent(action.__call__.__doc__,
                                            directive_indent)

diff --git a/source/sphinx_extensions/plugin_directory/templates/action.rst b/source/sphinx_extensions/plugin_directory/templates/action.rst
index 8b831c6..3e9d90c 100644
--- a/source/sphinx_extensions/plugin_directory/templates/action.rst
+++ b/source/sphinx_extensions/plugin_directory/templates/action.rst
@@ -11,7 +11,8 @@
       <div class="tab-content">
          <div id="cli" class="tab-pane fade in active">
             <pre>
-{{ cli_help }}   </pre>
+              {{- cli_help -}}
+            </pre>
          </div>
          <div id="api" class="tab-pane fade">
             <pre>
```

Now the markup looks nice:

screen shot 2017-09-26 at 9 16 31 pm

The rendered tab looks the exact same (no extra newlines):
screen shot 2017-09-26 at 9 26 23 pm

The rendered HTML:

screen shot 2017-09-26 at 9 25 17 pm

I'm not sure if it is worth changing the code in this PR though.

jairideout: This is perfect, thanks for the fix @thermokarst! I pushed up those changes.

jairideout:
Thanks for reviewing @thermokarst! Ready for final review (see squash commit).

PR #217, opened 2017-10-17
BUG: correct deblur doc

RNAer:
@wasade , this is it, right?

### DIFF
--- a/source/tutorials/moving-pictures.rst
+++ b/source/tutorials/moving-pictures.rst
@@ -159,8 +159,18 @@ Next, the Deblur workflow is applied using the ``qiime deblur denoise-16S`` meth
      --p-sample-stats \
      --o-stats deblur-stats.qza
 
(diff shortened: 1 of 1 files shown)


wasade:
Would it be possible to add an entry linking to the resulting `.qzv` and a light description of what is contained in there?

RNAer:
Just looked at it - I can't find in the rst source file the section of `output` rendered in html page. It seems programmatically generated?

jairideout:
[Check out the readme](https://github.com/qiime2/docs#contributing-to-the-docs), it describes the build targets. You can run `make html` to have the commands executed during build, which generates the `.qza`/`.qzv` files.

You'll probably want to add a command for generating a visualization of the deblur stats (I think that's what @wasade is asking for). This text should also get updated:

image

jairideout:
@RNAer or @wasade any progress on this?

PR #299, opened 2018-04-17
ENH: more citations in docs

ebolyen:
# Plugin
image

# Action
image
**(copied the registration, in practice plot wouldn't really have its own citations)**

### DIFF
--- a/source/sphinx_extensions/plugin_directory/templates/action.rst
+++ b/source/sphinx_extensions/plugin_directory/templates/action.rst
@@ -1,6 +1,28 @@
 {{ title }}
 {{ '=' * title|length }}
 
+{% if has_citations %}
+.. raw:: html
+
+   <table class="table action-info">
+     <tbody>
+       <tr>
+         <th scope="row"><p></p>Citations</th>
+         <td>
+
+.. bibliography:: {{ bib_id }}.bib
+   :all:
+   :keyprefix: {{ bib_id }}:
+   :labelprefix: {{ bib_id}}:
+
+.. raw:: html
+
+         </td>
+       </tr>
+     </tbody>
+   </table>
+{% endif %}
+
 .. raw:: html
 
    <div class="tabbed">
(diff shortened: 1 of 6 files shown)


source/sphinx_extensions/plugin_directory/templates/action.rst
```diff
@@ -1,6 +1,28 @@
 {{ title }}
 {{ '=' * title|length }}
 
+{% if has_citations %}
+.. raw:: html
+
+   <table class="table action-info">
+     <tbody>
+       <tr>
+         <th scope="row"><p></p>Citations</th>
```
ebolyen: The extra `<p></p>` is just because the sphinx extension adds an empty paragraph before the bibliography that I couldn't get rid of. So this makes the vertical alignment work again.

source/sphinx_extensions/plugin_directory/templates/plugin.rst
```diff
@@ -27,14 +27,24 @@
...
            {% endfor %}
+
+{% if plugin.citations %}
+.. raw:: html
+
          </td>
        </tr>
        <tr>
-         <th scope="row">Citation</th>
+         <th scope="row"><p></p>Citations</th>
```
ebolyen: Same here

source/theme/static/style.css
```diff
@@ -144,3 +144,16 @@ form.search input[name=q]{
   }
 
 }
+
+/* Fix up sphinxcontrib-bibtex html */
+table.citation {
+  border-left: none;
```
thermokarst: I toggled this rule and gosh, that border is so ugly!

source/theme/static/style.css
```diff
@@ -144,3 +144,16 @@ form.search input[name=q]{
...
+}
+
+table.citation td {
+  padding: 0px 0px 5px 0px;
+}
+
+table.citation td.label {
+  display: none;
```
thermokarst: 👍

source/sphinx_extensions/plugin_directory/templates/action.rst
thermokarst: 👍

ebolyen:
Sounds like a plan, thanks @thermokarst for all these reviews!

PR #314, opened 2018-06-13
DOC: Added metadata example

johnchase:
Currently the API tutorial mentions metadata, but the only use of it in
the tutorial is in a command line call. This should help clarify the
loading of and usage of metadata in the qiime2 API


johnchase:
I don't know if this would be useful for others, though I have had to remind myself a few times how to load metadata into qiime as it doesn't use the typical `Artifact.load()` method. It would be nice to have it referenced in the API tutorial

PR #326, opened 2018-07-05
BUG: added import and invocation to action and plugin pages

### LINKED ISSUE #307: call signature is needed in Artifact API action docs
The Artifact API docs for actions needs to show how the call signature for actions, similar to the usage line for the cli docs:

screenshot 2018-05-10 07 56 00

screenshot 2018-05-10 07 55 55

ChrisKeefe:
fixes #307

Two aesthetic questions for discussion:

- [x] Copy to clipboard:
It is possible to add copy-to-clipboard buttons to both plugin- and action-page pre tags. I opted not to do so, because they felt cluttered, and I'm not convinced that copy-buttons bring much value in this context. 

**Compare:** 

cropnocopy

cropcopybuttons

- [x] Duplication of CLI Docstring Usage line:
On Action pages, I wanted parity in the look of the CLI and API tabs, so I added an "Invocation" block to the CLI tab to mirror the "Import" block in the API tab. The new "Invocation" block duplicates the usage text in the docstring, which looks a bit sloppy to me. 

We can: 
- Merge as is, retaining the content duplication (result: screencaps below, pro: quick/easy, con: duplication)
- Remove the "Invocation" block from the CLI tab (pro: quick/easy, con: loss of parity)
- Remove the "Usage" line from the CLI docstring that is shown in these docs, with no change to the docstring shown in the CLI itself. (pro: deduplicates, preserves parity, con: additional effort/code)

API: 
cropactionnocopy

CLI: 
crop cli usage duplicated

### DIFF
--- a/source/sphinx_extensions/plugin_directory/extension.py
+++ b/source/sphinx_extensions/plugin_directory/extension.py
@@ -57,6 +57,7 @@ def generate_rst(app):
             write_bibtex(plugin.citations, index_bib)
 
         for action in plugin.actions.values():
+            action_api_name = action.id
             action_cli_name = action.id.replace('_', '-')
             action_path = os.path.join(plugin_dir, '%s.rst' % action_cli_name)
 
@@ -84,8 +85,12 @@ def generate_rst(app):
                 template = env.get_template('action.rst')
                 rendered = template.render(title=title, cli_help=cli_help,
                                            api_help=api_help, bib_id=bib_id,
-                                           has_citations=bool(action.citations)
-                                           )
+                                           has_citations=bool(
+                                            action.citations),
+                                           plugin_cli_name=plugin_cli_name,
+                                           action_cli_name=action_cli_name,
+                                           plugin_api_name=plugin.name,
+                                           action_api_name=action_api_name)
                 fh.write(rendered)
 
 
--- a/source/sphinx_extensions/plugin_directory/extension.py
+++ b/source/sphinx_extensions/plugin_directory/extension.py
@@ -88,8 +88,6 @@ def generate_rst(app):
                                            api_help=api_help, bib_id=bib_id,
                                            has_citations=bool(
                                             action.citations),
-                                           plugin_cli_name=plugin_cli_name,
-                                           action_cli_name=action_cli_name,
                                            import_path=import_path,
                                            action_api_name=action_api_name)
                 fh.write(rendered)
--- a/source/sphinx_extensions/plugin_directory/extension.py
+++ b/source/sphinx_extensions/plugin_directory/extension.py
@@ -57,7 +57,6 @@ def generate_rst(app):
             write_bibtex(plugin.citations, index_bib)
 
         for action in plugin.actions.values():
-            action_api_name = action.id
             action_cli_name = action.id.replace('_', '-')
             action_path = os.path.join(plugin_dir, '%s.rst' % action_cli_name)
 
@@ -90,7 +89,7 @@ def generate_rst(app):
                                            plugin_cli_name=plugin_cli_name,
                                            action_cli_name=action_cli_name,
                                            plugin_api_name=plugin.name,
-                                           action_api_name=action_api_name)
+                                           action_api_name=action.id)
                 fh.write(rendered)
 
 
--- a/source/sphinx_extensions/plugin_directory/extension.py
+++ b/source/sphinx_extensions/plugin_directory/extension.py
@@ -57,6 +57,8 @@ def generate_rst(app):
             write_bibtex(plugin.citations, index_bib)
 
         for action in plugin.actions.values():
+            import_path, action_api_name = action.get_import_path().rsplit(
+                '.', 1)
             action_cli_name = action.id.replace('_', '-')
             action_path = os.path.join(plugin_dir, '%s.rst' % action_cli_name)
 
@@ -88,8 +90,8 @@ def generate_rst(app):
                                             action.citations),
                                            plugin_cli_name=plugin_cli_name,
                                            action_cli_name=action_cli_name,
-                                           plugin_api_name=plugin.name,
-                                           action_api_name=action.id)
+                                           import_path=import_path,
+                                           action_api_name=action_api_name)
                 fh.write(rendered)
 
 
--- a/source/sphinx_extensions/plugin_directory/templates/action.rst
+++ b/source/sphinx_extensions/plugin_directory/templates/action.rst
@@ -40,7 +40,7 @@
 
 .. command-block::
 
-       from qiime2.plugins.{{plugin_api_name}}.actions import {{action_api_name}} as {{action_api_name}}
+       from {{import_path}} import {{action_api_name}}
 
 .. raw:: html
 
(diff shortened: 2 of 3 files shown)


thermokarst:
> It is possible to add copy-to-clipboard buttons to both plugin- and action-page pre tags. I opted not to do so, because they felt cluttered, and I'm not convinced that copy-buttons bring much value in this context.

Do the [changes proposed here](https://github.com/qiime2/docs/pull/327) make this feel less cluttered to you? I agree that their value isn't extraordinary in this context, but perhaps if they are less obtrusive they might be a welcome addition.

> Duplication of CLI Docstring Usage line:

I think the "Invocation" block is probably unnecessary, and I don't think parity between these pages is particularly valuable from a UX perspective (but I am probably wrong in that assessment). Curious to hear @ebolyen & @nbokulich's thoughts.

source/sphinx_extensions/plugin_directory/templates/action.rst
```diff
@@ -32,11 +32,15 @@
...
-           {{- cli_help -}}
-         </pre>
+         <h4>Invocation:</h4>
+         <pre>qiime {{plugin_cli_name}} {{action_cli_name}} [OPTIONS]</pre>
+         <h4>Docstring:</h4>
+         <pre>{{- cli_help -}}</pre>
        </div>
        <div id="api" class="tab-pane fade">
+       <h4>Import:</h4>
+       <pre>from qiime2.plugins.{{- plugin_api_name -}}.actions import {{action_api_name}} as {{action_api_name}}</pre>
```
thermokarst: `{{- plugin_api_name -}}` --- the `-`s tacked onto the mustaches indicate that jinja should [strip leading and trailing whitespace](http://jinja.pocoo.org/docs/2.10/templates/#whitespace-control) - is that necessary here? I wouldn't expect to see leading and trailing whitespace present in this value, but perhaps your development experience here indicated otherwise --- let me know!

source/sphinx_extensions/plugin_directory/extension.py
```diff
@@ -57,6 +57,7 @@ def generate_rst(app):
             write_bibtex(plugin.citations, index_bib)
 
         for action in plugin.actions.values():
+            action_api_name = action.id
```
thermokarst: This is super nitpicky, but maybe this makes more sense inlined into the context definition, [below](https://github.com/qiime2/docs/pull/326/files#diff-e380915e2a96fdaf7dd9aa48d1fb50d8R93) (like you did for `plugin_api_name`). Functionally there is no difference, but maybe keeping things consistent will help future readers of this code...

nbokulich:
>I think the "Invocation" block is probably unnecessary, and I don't think parity between these pages is particularly valuable from a UX perspective (but I am probably wrong in that assessment). Curious to hear @ebolyen & @nbokulich's thoughts.

I agree, drop CLI invocation — parity is not really essential in this context.

source/sphinx_extensions/plugin_directory/templates/action.rst
```diff
@@ -32,11 +32,19 @@
...
-         </pre>
+         <h4>Docstring:</h4>
+         <pre>{{- cli_help -}}</pre>
        </div>
        <div id="api" class="tab-pane fade">
+       <h4>Import:</h4>
+
+.. command-block::
+
+       from qiime2.plugins.{{plugin_api_name}}.actions import {{action_api_name}} as {{action_api_name}}
```
ebolyen: The last part of this is redundant:

```python
from qiime2.plugins.example_plugin.actions import example_action as example_action
```

is the same as just:

```python
from qiime2.plugins.example_plugin.actions import example_action
```

source/sphinx_extensions/plugin_directory/extension.py
```diff
@@ -84,8 +84,12 @@ def generate_rst(app):
...
                 rendered = template.render(title=title, cli_help=cli_help,
                                            api_help=api_help, bib_id=bib_id,
-                                           has_citations=bool(action.citations)
-                                           )
+                                           has_citations=bool(
+                                            action.citations),
+                                           plugin_cli_name=plugin_cli_name,
+                                           action_cli_name=action_cli_name,
+                                           plugin_api_name=plugin.name,
+                                           action_api_name=action.id)
```
ebolyen: Unfortunately this won't exactly work. The reason is that `plugin.name` will be something like `"feature-table"` whereas the Python API will see `feature_table`. You could use `.replace('-', '_')` but the better way to do this would be to call `action.get_import_path()` and then pop off the last segment.

source/sphinx_extensions/plugin_directory/templates/plugin.rst
```diff
@@ -27,6 +27,30 @@
...
+
+         </td>
+       </tr>
+       <tr>
+         <th scope="row">Artifact API Import</th>
+         <td>
+
+.. command-block::
+
+             from qiime2.plugins import {{ title }}
```
ebolyen: This has the same hyphenation problem as the import above. Unfortunately you will probably need to just call `.replace(...)` as popping off two "segments" of the import path requires about as much "internal" knowledge about the way the Artifact API is structured as replacing the hyphens.

Something that would make everyone's lives easier is if there was a `.id` property on plugins which matched the underscore replacement of the `.id` property on actions (and similarly a `get_import_path()` method).

ebolyen:
Sorry this review is so late!

source/sphinx_extensions/plugin_directory/templates/action.rst
ChrisKeefe: Sure is, @ebolyen. My thinking was that it gave users new to Python an easy opportunity to rename their import, but maybe this is too much like "teaching users Python"?

ebolyen: Unfortunately the more likely result would be that everyone would copy/paste the import with the redundant `as` expression. So I would leave it out, as others might also interpret the redundant expression as having some significance.

(Also, generally speaking, in Python people don't rename imports as a rule unless there's something to be gained from it (or its convention like `numpy` -> `np`).)

ChrisKeefe: Removed. Thanks!

source/sphinx_extensions/plugin_directory/extension.py
```diff
@@ -84,8 +86,12 @@ def generate_rst(app):
                 template = env.get_template('action.rst')
                 rendered = template.render(title=title, cli_help=cli_help,
                                            api_help=api_help, bib_id=bib_id,
-                                           has_citations=bool(action.citations)
-                                           )
+                                           has_citations=bool(
+                                            action.citations),
+                                           plugin_cli_name=plugin_cli_name,
+                                           action_cli_name=action_cli_name,
```
ebolyen: I don't think these two variables are used anymore: `plugin_cli_name` and `action_cli_name`.

ChrisKeefe: Sorry I missed that. Corrected. Thanks!

ebolyen:
Thanks @ChrisKeefe!

PR #336, opened 2018-07-19
BUG: fix python command-block to no-exec

ebolyen:
The command block directive is trying to run a shell command. Just add the `:no-exec:` directive argument to prevent that issue.

(Also fixed indentation)

### DIFF
--- a/source/sphinx_extensions/plugin_directory/templates/action.rst
+++ b/source/sphinx_extensions/plugin_directory/templates/action.rst
@@ -39,8 +39,9 @@
        <h4>Import:</h4>
 
 .. command-block::
+   :no-exec:
 
-       from {{import_path}} import {{action_api_name}}
+   from {{import_path}} import {{action_api_name}}
 
 .. raw:: html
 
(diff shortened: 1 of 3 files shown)


source/sphinx_extensions/plugin_directory/templates/action.rst
```diff
@@ -39,8 +39,9 @@
        <h4>Import:</h4>
 
 .. command-block::
+   :no-exec:
 
-       from {{import_path}} import {{action_api_name}}
+   from {{import_path}} import {{action_api_name}}
```
thermokarst: Can you drop the blank line below the import? @ChrisKeefe & I were just looking at this - it adds a newline to the copy-to-clipboard button.

ebolyen: Removing the newline does not seem to fix the issue (also all of our directives are formatted with an extra newline to separate the directive from the rest of the content).

thermokarst: Bummer.

ChrisKeefe:
Looks good to me, Evan. Thanks!

PR #337, opened 2018-07-19
BUG: One more no-exec on the rendered help text

thermokarst:
Tested locally with `make html`, this builds without error (well, tested up to the tutorials, then terminated the build).

### DIFF
--- a/source/sphinx_extensions/plugin_directory/templates/plugin.rst
+++ b/source/sphinx_extensions/plugin_directory/templates/plugin.rst
@@ -32,6 +32,7 @@
          <td>
 
 .. command-block::
+   :no-exec:
 
    qiime {{ title }}
 


thermokarst:
Oh wait, I just realized, #326 might not work quite as expected for plugins like q2-types:

screen shot 2018-07-19 at 4 26 54 pm

This is not valid:

```bash
qiime types
```

@ebolyen, thoughts?

ebolyen:
Eh, I think it's good enough for now. Let's worry about the invalid command another day. Maybe when interfaces can be asked about their representations?

thermokarst:
We could also test for any actions in the plugin, but that sounds like the job of an interface, not the docs.

ebolyen:
Exactly! Let's ignore for now.

PR #358, opened 2018-10-06
Reorganize and clarify import tutorial

cduvallet:
Okay, I re-ordered the sections, added a TOC, and clarified a few of the sections that have consistently confused me.

A couple of questions for y'all:

1. is there a way to import fasta files? i.e. not representative sequences but just quality-filtered fasta files, somehow associated with their sample IDs... this is obviously not ideal, but I find that many data sharers tend to only share fasta files so they're still more common than we'd like...
2. what happens if you have a rooted phylogenetic tree? can you import that?

Apologies for any typos/oversights, pretty sleepy tonight! 😪 😴


ebolyen:
Sorry, I've been super behind on notifications and @thermokarst is on vacation!

Re 1: I agree, this seems to happen fairly often. Really we just need someone to write a transformer or two to convert between demux FASTA and the single-file (but still demux) seqs.fna from QIIME 1. Until then, that isn't supported yet.

Re 2: Definitely, just change the type. I think that's something that our documentation could be much clearer about (apologies if you've addressed this already, I haven't actually looked yet). The commands are very much a black box right now, but if we explained what the format and type meant, then you could probably extrapolate some of the other imports without needing to enumerate all possible combinations.

ebolyen:
You're a hero @cduvallet!

I'm afraid not within the user docs, you'll be among the first to write about it. There is the [too technical dev docs](https://dev.qiime2.org/latest/storing-data/) which you could use as a reference for why things are the way they are, but it also doesn't have any practical explanation of how an end-user would use this.

cduvallet:
Ok, I made those edits and it should be ready to incorporate! (Though github is telling me there are conflicts that must be resolved... Do I need to do that or do y'all take care of it?)

That said, I think that these two issues we brought up are pretty big ones for QIIME 2 - what should I do to make sure that they're documented somewhere? Happy to open github issues in the right places, happy to let y'all manage that! To recap, they are:

1. There needs to be a way to import non-representative quality-filtered FASTA files which are somehow associated with their sample IDs.

2. The QIIME 2 docs really need an explanation of what each data type and data formats are (even just one sentence describing each one would make the world of difference!)

thermokarst:
Thanks so much @cduvallet!

> (Though github is telling me there are conflicts that must be resolved... Do I need to do that or do y'all take care of it?)

I am happy to wrangle that!

RE: point 1 - an issue in q2-types would be really helpful (it can cross-ref this issue for context).
RE: point 2 - we are with you 110%! I don't think we have any good issues floating out there, but here are the missing pieces that I see:

1. The framework's [base format definitions](https://github.com/qiime2/qiime2/tree/master/qiime2/plugin/model) will need to be updated to support some kind of help or description text.
2. Prose will need to be written for all of the formats. Most of them are defined in q2-types.
3. Interfaces and docs will need to be made aware of this new format attribute so that they can be rendered out.

That is the "hard way" approach - a more useful short-term solution could be a forum post or user doc resource that just includes the prose (like `FormatXYZ` - this is for blah blah blah, and looks like the following example: XYZ). Maybe that could be the initial target - that would allow us to port the prose to the formats whenever the machinery is in place...

I can take point on writing up issues for the format description steps, if that sounds okay to you.

Thanks! 

PS - I will start reviewing this PR shortly - so sorry this has been held up on our end for a while now, thank you for your patience!

source/tutorials/importing.rst
```diff
@@ -66,10 +82,18 @@ Importing data
 Format description
 ******************
 
-In the "Earth Microbiome Project (EMP) protocol" format for paired-end reads, there are three ``fastq.gz`` files, one containing forward sequence reads, one containing reverse sequence reads, and one containing the associated barcode reads, with the sequence data still multiplexed. The order of the records in the three ``fastq.gz`` files defines the association between the sequences reads and barcode reads.
+If your FASTQ files follow the "`Earth Microbiome Project (EMP) protocol`_" format for `paired-end` reads, you should have three ``fastq.gz`` files total:
```
thermokarst:
```suggestion
Paired-end "`Earth Microbiome Project (EMP) protocol`_" formatted reads should have three ``fastq.gz`` files total:
```

source/tutorials/importing.rst
```diff
@@ -97,19 +121,94 @@ Importing data
...
+     --input-path casava-18-paired-end-demultiplexed \
+     --input-format CasavaOneEightSingleLanePerSampleDirFmt \
+     --output-path demux-paired-end.qza
+
 .. _`manifest file`:
 
 "Fastq manifest" formats
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
+If you don't have either EMP or Casava format, you need to import your data into QIIME 2 manually.
```
thermokarst: How about:

```suggestion
If you don't have either EMP or Casava format, you need to import your data into QIIME 2 manually using a "manifest" format.
```
?

source/tutorials/importing.rst
```diff
@@ -97,19 +121,94 @@ Importing data
...
 
-For single-end reads, there must be exactly one line per sample id corresponding to either the forward or reverse reads. For paired-end reads there must be exactly two lines per sample id, corresponding to the forward and the reverse reads. The direction field on each line can only contain the text ``forward`` or ``reverse``.
+For single-end reads, there must be exactly one line per sample id corresponding to the forward reads (or reverse, if you're processing these separately as single-end reads). For paired-end reads there must be exactly two lines per sample id, corresponding to the forward and the reverse reads. The direction field on each line can only contain the text ``forward`` or ``reverse``.
```
thermokarst: Minor style change:

```suggestion
For single-end reads, there must be exactly one line per sample ID corresponding to the forward reads (or reverse, if you're processing these separately as single-end reads). For paired-end reads there must be exactly two lines per sample id, corresponding to the forward and the reverse reads. The direction field on each line can only contain the text ``forward`` or ``reverse``.
```

source/tutorials/importing.rst
```diff
@@ -123,37 +222,37 @@ The ``fastq.gz`` absolute filepaths may contain environment variables (e.g., ``$
...
+Just like with ``fastq.gz``, the absolute filepaths in the manifest for any ``fastq`` files must be accurate. The following example illustrates a simple fastq manifest file for ``fastq`` single-end read data for one sample.
 
 ::
 
   sample-id,absolute-filepath,direction
   sample-1,$PWD/some/filepath/sample1_R1.fastq,forward
-  sample-1,$PWD/some/filepath/sample1_R2.fastq,reverse
 
-There are four variants of this format which are defined in the following sections.
+There are four variants of FASTQ data which you must specific to QIIME 2 when importing, and which are defined in the following sections.
```
thermokarst: Typo:

```suggestion
There are four variants of FASTQ data which you must specify to QIIME 2 when importing, and which are defined in the following sections.
```

source/tutorials/importing.rst
```diff
@@ -123,37 +222,37 @@ The ``fastq.gz`` absolute filepaths may contain environment variables (e.g., ``$
   sample-1,$PWD/some/filepath/sample1_R2.fastq.gz,reverse
   sample-2,$PWD/some/filepath/sample2_R2.fastq.gz,reverse
 
-Just like with ``fastq.gz``, the absolute filepaths in the manifest for any ``fastq`` files must be accurate. The following example illustrates a simple fastq manifest file for ``fastq`` paired-end read data for one sample.
+Just like with ``fastq.gz``, the absolute filepaths in the manifest for any ``fastq`` files must be accurate. The following example illustrates a simple fastq manifest file for ``fastq`` single-end read data for one sample.
```
thermokarst: What if we made it two samples? Then there is a bit of symmetry with the psuedo-example above.

source/tutorials/importing.rst
cduvallet: Sure, though I find the focus on calling "manifest" a format very confusing. Maybe because the "manifest" format has something called a manifest file, but the manifest file itself is not an actual format (it's a file). I can see where this terminology comes from if I try to explain it from a CS point of view, but as a computational biologist it was very confusing to see the same word used for a format and a file.

But keeping things consistent with the rest of the QIIME 2 docs and code is more important than my confusion, so that's just a "for what it's worth!" :)

source/tutorials/importing.rst
cduvallet: Yes, I also found this confusing upon first read. That said, I do think it's useful to have an example of single-end reads as well (I can imagine novice users wondering "what do I put for the reverse read if I have single-end reads?)

So I don't feel strongly either way!

source/tutorials/importing.rst
```diff
@@ -33,10 +42,17 @@ Sequence data with sequence quality information
...
+1. one ``fastq.gz`` file that contains the single-end reads,
+2. and another that contains the associated barcode reads
+
+In this format, sequence data is still multiplexed (i.e. you have only one ``fastq.gz`` file containing raw data for all of your samples.
+
+The order of the records in the two ``fastq.gz`` files defines the association between a sequence read and its barcode read (i.e. the first barcode read corresponds to the first sequence read, the second barcode to the second read, and so on.)
 
 Obtaining example data
-**********************
+```````````````````````
```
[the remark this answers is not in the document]
cduvallet: 🎉 

After I offered to re-organize this tutorial, I realized that like 40% of my confusion was actually just being caused by the different header levels. Unfortunately rst format didn't let me make them _all_ perfectly consistent (e.g. in sections with different numbers of sub-heading levels, I couldn't skip an unused level directly to a lower level), but hopefully this helps a little!

cduvallet:
> RE: point 1 - an issue in q2-types would be really helpful (it can cross-ref this issue for context).

Okay, will do that now!

> RE: point 2 - we are with you 110%! 

🎉 

> Maybe that could be the initial target - that would allow us to port the prose to the formats whenever the machinery is in place... I can take point on writing up issues for the format description steps, if that sounds okay to you.

That sounds great @thermokarst! When those are finished, it will be crucial to link to them in this tutorial too. 

> Hey there @cduvallet! Made it halfway through this but have to head out the door. I love these changes, and am really excited for this! Sorry it took us so long to look at - things have been busy with the QIIME 2 manuscript plus our workshop travel schedule. Anyway, please stay tuned for part two some time on Monday. Have a great weekend!

Great, no worries! The delay between updates is quite nice actually - gives me time to work on other things guilt-free ;) Since the deadline for PRs for next release is on Tuesday, I'm guessing that this will wait till the next release?

source/tutorials/importing.rst
thermokarst: > though I find the focus on calling "manifest" a format very confusing

Oops, sorry, wasn't trying to emphasize that with my suggestion. I guess when I first read it I was thinking, "well, a new user might see this and think they aren't able to use QIIME 2 with their demuxed data," so my suggestion was coming from a place of clarifying that "yep, we can handle that."

I am okay with this being left off (as in, as it was before), I think on second read it is pretty darn clear that this is a statement within the context of the manifest formats section --- I will gladly go with whatever you decide to do.

source/tutorials/importing.rst
```diff
@@ -198,73 +297,71 @@ Importing Data
...
      --input-format PairedEndFastqManifestPhred64
 
+Sequences without quality information (i.e. FASTA)
+--------------------------------------------------------
+
 .. _`casava import`:
 
-Casava 1.8 single-end demultiplexed fastq
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Unfortunately, there is currently no way to import data files which contain quality-filtered FASTA sequences associated with sample IDs.
```
thermokarst: How about the [`QIIME1DemuxFormat`](https://github.com/qiime2/q2-types/blob/master/q2_types/per_sample_sequences/_format.py#L242-L273)? We have a tutorial that has a brief usage example: https://docs.qiime2.org/2018.8/tutorials/otu-clustering/

I could drop in a brief example of this format here in this section if that helps.

source/tutorials/importing.rst
thermokarst: Oh, good point --- that makes sense to me. How about we leave as-is?

source/tutorials/importing.rst
```diff
@@ -198,73 +297,71 @@ Importing Data
...
+
+Per-feature unaligned sequence data (i.e., representative FASTA sequences)
+--------------------------------------------------------------------------
 
 Format description
-******************
+~~~~~~~~~~~~~~~~~~
 
-In this format, there is one ``fastq.gz`` file for each sample in the study, and the file name includes the sample identifier. The file name for a single sample might look like ``L2S357_15_L001_R1_001.fastq.gz``. The underscore-separated fields in this file name are the sample identifier, the barcode sequence or a barcode identifier, the lane number, the read number, and the set number.
+Unaligned sequence data is imported from a fasta formatted file containing DNA sequences that are not aligned (i.e., do not contain `-` or `.` characters). The sequences may contain degenerate nucleotide characters, such as ``N``, but some QIIME 2 actions may not support these characters. See the `scikit-bio fasta format description`_ for more information about the fasta format.
```
thermokarst:
```suggestion
Unaligned sequence data is imported from a FASTA formatted file containing DNA sequences that are not aligned (i.e., do not contain `-` or `.` characters). The sequences may contain degenerate nucleotide characters, such as ``N``, but some QIIME 2 actions may not support these characters. See the `scikit-bio fasta format description`_ for more information about the FASTA format.
```

source/tutorials/importing.rst
```diff
@@ -198,73 +297,71 @@ Importing Data
...
 **********************
 
 .. download::
-   :url: https://data.qiime2.org/2018.11/tutorials/importing/casava-18-single-end-demultiplexed.zip
-   :saveas: casava-18-single-end-demultiplexed.zip
-
-.. command-block::
-
-   unzip -q casava-18-single-end-demultiplexed.zip
+   :url: https://data.qiime2.org/2018.8/tutorials/importing/sequences.fna
```
thermokarst:
```suggestion
   :url: https://data.qiime2.org/2018.11/tutorials/importing/sequences.fna
```

source/tutorials/importing.rst
```diff
@@ -198,73 +297,71 @@ Importing Data
...
 **********************
 
 .. download::
-   :url: https://data.qiime2.org/2018.11/tutorials/importing/casava-18-paired-end-demultiplexed.zip
-   :saveas: casava-18-paired-end-demultiplexed.zip
-
-.. command-block::
-
-   unzip -q casava-18-paired-end-demultiplexed.zip
+   :url: https://data.qiime2.org/2018.8/tutorials/importing/aligned-sequences.fna
```
thermokarst:
```suggestion
   :url: https://data.qiime2.org/2018.11/tutorials/importing/aligned-sequences.fna
```

source/tutorials/importing.rst
```diff
@@ -317,80 +414,57 @@ Importing data
...
+Phylogenetic trees are imported from newick formatted files. See the `scikit-bio newick format description`_ for more information about the newick format.
 
 Obtaining example data
-~~~~~~~~~~~~~~~~~~~~~~
+**********************
 
 .. download::
-   :url: https://data.qiime2.org/2018.11/tutorials/importing/sequences.fna
-   :saveas: sequences.fna
+   :url: https://data.qiime2.org/2018.8/tutorials/importing/unrooted-tree.tre
```
thermokarst:
```suggestion
   :url: https://data.qiime2.org/2018.11/tutorials/importing/unrooted-tree.tre
```

source/tutorials/importing.rst
```diff
@@ -97,19 +121,94 @@ Importing data
...
+2. the barcode sequence or a barcode identifier,
+3. the lane number,
+4. the direction of the read (i.e. only R1, because these are single-end reads), and
+5. the set number.
+
+Obtaining example data
+``````````````````````
+
+.. download::
+   :url: https://data.qiime2.org/2018.8/tutorials/importing/casava-18-single-end-demultiplexed.zip
```
thermokarst:
```suggestion
   :url: https://data.qiime2.org/2018.11/tutorials/importing/casava-18-single-end-demultiplexed.zip
```

source/tutorials/importing.rst
```diff
@@ -97,19 +121,94 @@ Importing data
...
+2. the barcode sequence or a barcode identifier,
+3. the lane number,
+4. the direction of the read (i.e. R1 or R2), and
+5. the set number.
+
+Obtaining example data
+``````````````````````
+
+.. download::
+   :url: https://data.qiime2.org/2018.8/tutorials/importing/casava-18-paired-end-demultiplexed.zip
```
thermokarst:
```suggestion
   :url: https://data.qiime2.org/2018.11/tutorials/importing/casava-18-paired-end-demultiplexed.zip
```

source/tutorials/importing.rst
```diff
@@ -198,73 +297,71 @@ Importing Data
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Per-feature aligned sequence data (i.e., aligned representative FASTA sequences)
+--------------------------------------------------------------------------------
 
 Format description
-******************
+~~~~~~~~~~~~~~~~~~
 
-In this format, there are two ``fastq.gz`` file for each sample in the study, and the file name includes the sample identifier. The forward and reverse read file names for a single sample might look like ``L2S357_15_L001_R1_001.fastq.gz`` and ``L2S357_15_L001_R2_001.fastq.gz``, respectively. The underscore-separated fields in this file name are the sample identifier, the barcode sequence or a barcode identifier, the lane number, the read number, and the set number.
+Aligned sequence data is imported from a fasta formatted file containing DNA sequences that are aligned to one another. All aligned sequences must be exactly the same length. The sequences may contain degenerate nucleotide characters, such as ``N``, but some QIIME 2 actions may not support these characters. See the `scikit-bio fasta format description`_ for more information about the fasta format.
```
thermokarst:
```suggestion
Aligned sequence data is imported from a FASTA formatted file containing DNA sequences that are aligned to one another. All aligned sequences must be exactly the same length. The sequences may contain degenerate nucleotide characters, such as ``N``, but some QIIME 2 actions may not support these characters. See the `scikit-bio fasta format description`_ for more information about the FASTA format.
```

source/tutorials/importing.rst
cduvallet: Got it, just made a change that should hopefully address both our concerns! :)

New text:

> If you don't have either EMP or Casava format, you need to import your data into QIIME 2 manually by first creating a "manifest file" and then using the `qiime tools import` command with a few more flags specified than in the EMP or Casava import commands.

source/tutorials/importing.rst
cduvallet: Oh, so this format does exist already? Does this mean [the issue (#197) I filed in q2-types](https://github.com/qiime2/q2-types/issues/197) isn't necessary?

If that's the case, then yes having you put a brief example here would be great!

source/tutorials/importing.rst
cduvallet: Sure!

cduvallet:
Ok, apart from the QIIME1DemuxFormat example, I think it should be good to go!

source/tutorials/importing.rst
thermokarst: No, I think that issue makes sense to me, since we don't currently support importing per-sample demuxed FASTA. The `QIIME1DemuxFormat` is a single-file demuxed format that uses a weird header syntax. It is great for QIIME 1 data, but I haven't really seen too many other tools that use it.

cduvallet: ah ok, I see. Yeah, some way to specify these differences might be useful, and to give users at least one option to import FASTA sequences (even if it's a pain to format the sequences in this way...)

thermokarst:
This looks great @cduvallet! If [my changes](https://github.com/qiime2/docs/pull/358/commits/c126cb82358df9cc0ee46816d151f552b06fa656) look okay to you please let me know, otherwise amend away! Thanks again!

cduvallet:
Looks good! Added one sentence clarifying that other FASTA formats aren't supported, so that people don't try to go down those rabbit holes.

But other than that, looks good to me! 😄

thermokarst:
Good catch, I totally forgot about that! Okay, time to :shipit:! Thanks so much @cduvallet!

PR #363, opened 2018-11-08
update to feature-classifier tutorial: extract-reads length filtering

nbokulich:
Describe new `min-length` and `max-length` parameters in tutorial.


source/tutorials/feature-classifier.rst
```diff
@@ -64,12 +64,16 @@ It has been shown that taxonomic classification accuracy of 16S rRNA gene sequen
...
 
+.. note:: The example command above uses the ``min-length`` and ``max-length`` parameters to exclude simulated amplicons that are far outside of the anticipated length distribution using those primers. Such amplicons are likely non-target hits and should be excluded. If you adapt this command for your own use, be sure to select settings that are appropriate for the marker gene, no the settings used here. The ``min-length`` parameter is applied _after_ the ``trim-left`` and ``trunc-len`` parameters, and ``max-length`` _before_, so be sure to set appropriate settings to prevent valid sequences from being filtered out.
```
thermokarst:
```suggestion
.. note:: The example command above uses the ``min-length`` and ``max-length`` parameters to exclude simulated amplicons that are far outside of the anticipated length distribution using those primers. Such amplicons are likely non-target hits and should be excluded. If you adapt this command for your own use, be sure to select settings that are appropriate for the marker gene, not the settings used here. The ``min-length`` parameter is applied _after_ the ``trim-left`` and ``trunc-len`` parameters, and ``max-length`` _before_, so be sure to set appropriate settings to prevent valid sequences from being filtered out.
```

no -> not typo

PR #372, opened 2018-12-04
Copypasta error on beta-group-significance

wasade:
Beta group significance results from the alpha/beta core diversity section are identical but the output filenames suggest otherwise. On closer inspection, it looks as though the filename was updated in one of the commands but the specific metadata category used did not get updated.


wasade:
Separate from this, I just saw that some URLs such as the classifier for sklearn refer to 2018.8, should those be updated?

thermokarst:
> I just saw that some URLs such as the classifier for sklearn refer to 2018.8, should those be updated?

It wouldn't hurt (we start each release cycle with a find-and-replace for all old release numbers to the newest release --- this would've been updated if this branch was in `master`).

PR #379, opened 2019-01-08
set random seeds in q2-sample-classifier and q2-longitudinal tutorials

### LINKED ISSUE #366: sample-classifier tutorial: set random states and/or stratify
**Improvement Description**
Set random states in sample classifier tutorial, and stratify where relevant. Also pass through other tutorials (q2-longitudinal) and set on any ML methods.

**Current Behavior**
Lack of stratification might lead to stochastic errors with some datasets — this really just seems to be an issue with `maturity-index` (already fixed), but by setting a random state we ensure that we won't get any surprises.

nbokulich:
now all commands have random seeds specified.

fixes #366 

Also set a different seed for q2-feature-volatility — the original command ran but was making an excessively large viz that would not display volatility plots on some machines.

### DIFF
--- a/source/tutorials/longitudinal.rst
+++ b/source/tutorials/longitudinal.rst
@@ -339,7 +339,7 @@ This pipeline produces several output files:
 
 8. ``model_summary.qzv`` contains a summary of the model parameters used by the supervised learning estimator, as described in the :doc:`sample classifier tutorial <sample-classifier>` for the equivalently named outputs from the ``classify-samples`` pipeline.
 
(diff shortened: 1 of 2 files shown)


source/tutorials/longitudinal.rst
```diff
@@ -275,7 +275,7 @@ Let's test this out on the ECAM dataset. First download a table to work with:
      --p-state-column month \
      --p-individual-id-column studyid \
      --p-n-estimators 10 \
-     --p-random-state 123 \
+     --p-random-state 17 \
```
ChrisKeefe: Just curious - why the change away from `123` here, when we are setting `random-state` to 123 for all sample-classifier methods? And is there a reason we've set `maturity-index`'s `random-state` differently from the other passed values (i.e. `1010101`)?

PR #397, opened 2019-04-02
MAINT: Sphinx 2.0

thermokarst:
Sphinx 2.0 recently landed --- this changeset updates the user docs to work with the latest (and greatest). Most of the changes have to do with our custom admonitions, and some deprecated APIs. Also, I had to update the bibtex render style to something other than the default, which is a set of paragraph elements. Paragraphs get cross-references in Sphinx --- this causes the build to fail because there are no corresponding citations in the docs anywhere. My solution was to change to a bullet list, since those elements don't get the cross-reference treatment!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,4 @@
 # Documentation building requirements.
-# TODO: unpin this
-Sphinx==1.4.6
+Sphinx
 Jinja2
-# TODO: unpin this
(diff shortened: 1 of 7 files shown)


thermokarst:
Going to merge this as-is and will double-check the rendered docs that busywork produces.

Issue #402, opened 2019-04-28
New 16s Tutorial for Workshop


jwdebelius:
**Improvement Description**
@nbokulich and @thermokarst and I were discussing an alternative to the moving pictures tutorial for an upcoming QIIME 2 workshop in June. The goal would be have something slightly different and a little bit more modern

**Proposed Behavior**
I've got a fairly good start on a tutorial with 48 mouse samples from a 2016 paper and would be interested in contributing it here as a tutorial. It's quite similar to the moving pictures tutorial in terms of text/content. The current version covers:

1. Hypothesis based on a real experiment with cross sectional and longitudinal aspects
2. Metadata
2. Data loading via a manifest
3. Clustering with Deblur (~5 min)
4. Tree building via fragment insertion (11 minutes... longest run time)
5. Taxonomic classification 
6. Core diversity
    * Alpha group significance
    * Beta group significance
    * Adonis on a multivariate intersection
7. Taxonomy barchart
8. Adonis
9. Longitudinal analysis

nbokulich:
Sounds good!

See also the `summarize-*` methods in `q2-demux`. One thing we have discussed doing more in core tutorials (i.e. those we use for workshops) is to post multiple different subsamples of the data (e.g., see [here](https://github.com/qiime2/docs/blob/0f23a659fc0b828c73fc97f6529c6987f08ae22b/source/tutorials/fmt-cdiff.rst)), so that we can do things like demux, denoise, and build phylogenies (slow steps) with small datasets in workshops, but have the full dataset available (and pre-processed) for the downstream analysis.

Your dataset does not sound too big — but it may be worth seeing if a 10% sequence subset contains sufficient data for analysis.

jwdebelius:
I've got summarise.

Since there's interest, I'll put in a PR shortly (I just need to test building the docs, always the scariest for me!) But, the "cooking show" approach is probably a good option for faster run times. So, maybe subsample on the full table for running the data, and then have the "full" dataset somewhere else?

I did already subsample the data down to 3-5K sequences in an attempt to limit run time on the table, which produces a final depth of something like 3K seqs/sample max and comes out with a nice rarefaction depth of 1K.

nbokulich:
sorry, I didn't mean summarize (😪) I meant `subsample-*` ! Edited above. Thanks for understanding me nonetheless.

[...]

PR #403, opened 2019-04-29
 Parkinson's Mouse Tutorial. 

jwdebelius:
My PD mice tutorial from #402. An updated introduction tutorial with a more modern dataset.

I'm currently hosting the data on google drive and in a gist here which seems to be messing with the download and execution. I'm not sure how to best handle this.


jwdebelius:
I'm still not sure about data hosting and automated testing. Is there a temporary location to put the sequence files that would work?

thermokarst:
We have an AWS account to host any tutorial data on. If you want, I can upload and share links with you, just let me know.

jwdebelius:
@thermokarst, That would be great. Sorry, I've not done the AWS tutorial uploading before and the google drive solution does not seem to be working. I'll send you files.

thermokarst:
Data files:

- https://data.qiime2.org/2019.7/tutorials/pd-mice/demultiplexed_seqs.zip (I repackaged, removing your personally-identifying data that was zipped in there (thanks macos!))
- https://data.qiime2.org/2019.7/tutorials/pd-mice/manifest (converted to TSV, the new manifest format)
- https://data.qiime2.org/2019.7/tutorials/pd-mice/sample_metadata (for viewing)
- https://data.qiime2.org/2019.7/tutorials/pd-mice/sample_metadata.tsv (for wgeting)

Thanks for the patience!

You could also use your sample metadata for the manifest --- your call.

jwdebelius:
Thanks @thermokarst !

I tested the links with the latest qiime release, which seemed like it was 2019.4 and got an error in the build with the new tsv manifest format. Is this for the 2019.7 release. I've temporarily put back in the path to my csv gist to see if I can get it to build. 

Do we plan to teach the workshop off the new release or 2019.4?

Second, I'm getting errors off the mapping file since the underscores were replaced with dashes. Is it possible to go back to the older format? Ive got `sample_name` as my column identifier (since its data from qiita). It may be my bias, but I find snake_case easier. Plus, it's less proof reading?

jwdebelius:
It turns out i may have pulled two copies of the latest yaml and not cleaned my file space appropriately :blush:, so the tsv worked with that correction. Thanks for being patient with me, there.

I brought in your edits. I'm having an issue with kebab case for the metadata:

```bash
qiime longitudinal linear-mixed-effects \
  --m-metadata-file ./metadata.tsv \
  --m-metadata-file ./from_first_unifrac.qza \
  --p-metric Distance \
  --p-state-column 'days-post-transplant' \
  --p-individual-id-column mouse-id \
  --p-group-columns genotype,donor \
  --o-visualization ./from_first_unifrac_lme.qzv \
  --verbose
```
Thows an error:

```
Plugin error from longitudinal:

  Error evaluating factor: NameError: name 'days' is not defined
      Distance ~ days-post-transplant * genotype * donor
                 ^^^^
```
since - is a special character in patsy

thermokarst:
Glad to hear you got the manifest working. RE metadata, I reverted back to snake case (sorry I didn't make that more clear, above), so you can stick with that. If you did want to go with kebab case, you would need to quote the parameter in patsy (I just double checked, looks like this is supported).

jwdebelius:
It built! It looks good, so, I think its ready for a full review (or fuller review?)

jwdebelius:
Sorry @thermokarst, can I ping again on this if we're planning on using it for the workshop?

thermokarst:
Yep, sorry, I thought we had agreed on that weeks ago.

thermokarst:
@jwdebelius, I sent you an email about workshop logistics, maybe we can move the discussion over there?

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+
+Therefore, a gnotobiotic mouse study was utilized to evaluate the role of the microbiome in the development of PD symptoms. Feces were collected from six donors with Parkinson’s disease and six age- and sex-matched neurologically health controls, and then transplanted into mice who were either predisposed to developing Parkinson’s disease due to a mutation ("aSyn") or resistant wild type mice ("BDF1"). Mice from different donors were kept in seperate cages, but mix from different genetic backgrounds were co-housed. The mice were followed for 7 weeks to see if they developed symptoms of Parkinson’s disease.
+
+We’ll look a subset of data from two human donors (one healthy and one with PD) whose samples were transplanted into two cages of mice from the susceptible genotype.
+
+For this tutorial, a subset of the metadata has been pulled from the full file, and the sequences also been subsamples down to around 5000 sequences per sample from a much larger distribution to make the dataset easier to run in a short time. The sequences for the full study were deposited in EBI with accession `PRJEB17694`_; processed tables can be downloaded from the `qiita`_  database from study 10483.
```
thermokarst:
```suggestion
For this tutorial, a subset of the metadata has been pulled from the full file, and the sequences have also been subsampled down to around 5000 sequences per sample from a much larger distribution to allow the tutorial to run in a short time. The sequences for the full study were deposited in EBI with accession `PRJEB17694`_; processed tables can be downloaded from the `qiita`_  database from study 10483.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
+Parkinson’s Mouse Tutorial
+------------------------------------
+
+As an example of how to process data through QIIME 2, we will use a set of humanized murine fecal samples from `Sampson et al, 2016`_.
```
thermokarst: How about:

```suggestion
This tutorial will demonstrate a "typical" QIIME 2 analysis, using samples of humanized murine fecal samples from `Sampson et al, 2016`_.
```

thermokarst: Also, perhaps this sentence can just get merged with the next paragraph down (line 6)?

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+.. command-block::
+   :no-exec:
+
+   mkdir ./mouse_tutorial
+   cd ./mouse_tutorial
+
+Metadata
+========
+
+Before starting any analysis, it's important to be familiar with the metadata. In this study, the metadata file contains 7 columns.
```
thermokarst: > it's important to be familiar with the metadata

:heart:  - love this

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+|                         | healthy)           |                 |                  |
++-------------------------+--------------------+-----------------+------------------+
+| days_post_transplant    | the number of      | numeric         | 7, 14, 21, 49    |
+|                         | days after the     |                 |                  |
+|                         | mice were          |                 |                  |
+|                         | humanized          |                 |                  |
++-------------------------+--------------------+-----------------+------------------+
+
+
+Even though the mouse ID looks like a number, we will specify the type using the ``#q2_type`` column in the data.
```
thermokarst: This is awesome, will be really useful in workshop situations, for discussing the column type directives.

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+   :saveas: metadata.tsv
+
+The sample metadata will be used through out the tutorial.
+
+Loading the data into QIIME 2
+=============================
+
+In QIIME 2, all data is structured as an Artifact of a specific semantic type. The artifacts contain the data as well as information about the data, including a record of the original data, the tools used to process it. This allows for better tracking of how you actually got to where you are in your analysis. You can learn more about common QIIME 2 Artifacts and types of artifacts :doc:`here <../semantic-types/>`.
+
+Our samples were amplified u|sing the `EMP 515f-806r`_ primers and sequenced on an Illumina MiSeq with a 2x150bp kit. The hypervariable
```
thermokarst:
```suggestion
Our samples were amplified using the `EMP 515f-806r`_ primers and sequenced on an Illumina MiSeq with a 2x150bp kit. The hypervariable
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+
+We will load the sequences as ``SampleData[SequencesWithQuality]``, which is the single end sequence demultiplexed format. If we wanted to import paired sequences, we would chose the ``SampleData[PairedEndSequencesWithQuality]`` type. We will import the sequences using the sample manifest format. This is one of the most versatile ways to import demultiplexed data in QIIME 2. We create a tab-separated sample manifest file that maps the sample name we want to use in QIIME 2 to the path to the sequence file, and the read direction. The benefit is that the demultiplexed sequence files can be named anything you want; there are not fixed assumptions about the conventions, and the file names do not dictate the final name. When QIIME 2 reads the file, it ignores any line prefixed with the ``#`` symbol. The first line that doesn’t contain a ``#`` is the header line and must be ``sample-id\tabsolute-filepath``. The sample order after the header line does not matter.
```
thermokarst:
```suggestion
We will import the sequences as ``SampleData[SequencesWithQuality]``, which is the single end sequence demultiplexed format. If we wanted to import paired sequences, we would choose the ``SampleData[PairedEndSequencesWithQuality]`` type. We will import the sequences using the sample manifest format. This is one of the most versatile ways to import demultiplexed data in QIIME 2. We create a tab-separated sample manifest file that maps the sample name we want to use in QIIME 2 to the path to the sequence file. The benefit is that the demultiplexed sequence files can be named anything you want; there are not fixed assumptions about the conventions, and the file names do not dictate the final name. When QIIME 2 reads the file, it ignores any line prefixed with the ``#`` symbol. The first line that doesn’t contain a ``#`` is the header line and must be ``sample-id\tabsolute-filepath``. The sample order after the header line does not matter.
```

I wonder if we should just add this column to the sample metadata file and call it a day. Thoughts @jwdebelius & @nbokulich?

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+   unzip demuliplexed_seqs.zip
+
+You can use the ``head`` command to check the first five lines of the sample manifest.
+
+.. command-block::
+   :no-exec:
+
+   head -n 6 manifest
+
+When using the single end manifest format, a sample name can only appear in one line and can only map to one sequencing file. In a paired end manifest, the sample name should appear twice. The **absolute-filepath** for each sample must be an `absolute path`_, which specifies the “full” location of the file. We do that here using the ``$PWD`` variable, which uses the local absolute directory. The **direction** column can only have two values: *forward* and *reverse*.
```
thermokarst:
```suggestion
When using the manifest formats, a sample name can only appear in one line and can only map to one sequencing file per column (one column for single-end, two columns for paired-end). The **absolute-filepath** for each sample must be an `absolute path`_, which specifies the "full" location of the file. We do that here using the ``$PWD`` variable, which uses the local absolute directory.
```

Updating this text to match the spec for the V2 manifest formats.

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+Let’s check the sequences and the sequencing depth of the samples using the ``qiime demux summarize`` command. It provides information about the number of sequences in each sample, as well as the quality of the sequences.
+
+Before running the command, let’s review the help documentation to make sure we understand the arguments.
+
+.. command-block::
+   :no-exec:
+
+   qiime demux summarize --help
+
+Based on the documentation, we should pass the demultiplexed sequences that we imported as the ``--i-data`` argument, since this takes a ``SequencesWithQuality]`` semantic type, and that’s the type of data we imported. We’ll specify the location we want the visualization by passing the output path to ``--o-visualization``. However, to speed up the command process, we’ll change the ``--p-n`` parameter to 1000. This means that rather than resampling the sequences 10000 times (the default number) to get the quality score, we’ll only re-sample them 1000 times.
```
thermokarst:
```suggestion
Based on the documentation, we should pass the demultiplexed sequences that we imported as the ``--i-data`` argument, since this takes a ``SampleData[SequencesWithQuality`` semantic type, and that’s the type of data we just imported. We’ll specify the location we want to save the visualization to by specifying the output path to ``--o-visualization``. However, to speed up the command process, we’ll change the ``--p-n`` parameter to 1000. This means that rather than resampling the sequences 10000 times (the default number) to get the quality score, we’ll only re-sample them 1000 times.
```

Not sure about changing the default to 1000, seems like the difference in runtime is pretty negligible. What do you think, @jwdebelius?

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+.. command-block::
+
+   qiime demux summarize \
+     --i-data ./demux_seqs.qza \
+     --p-n 1000 \
+     --o-visualization ./demux_seqs.qzv
+
+You can view the .qzv visualization file at `view.qiime2.org`_. Just drag and drop the file into the viewer window.
+
+.. question::
```
thermokarst: :heart_eyes: Great questions, especially in a workshop environment.

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+
+ASVs are a more recent development and provide better resolution in features than traditional OTU-based methods. ASVs can separate features based on differences of a single nucleotide in sequences of 400 bp or more, a resolution not possibly even with 99% identity OTU clustering. QIIME 2 currently offers denoising via `Dada2`_ (``q2-dada2``) and `Deblur`_ (``q2-deblur``). The major differences in the algorithms and motivation for denoising are nicely described in `Nearing et al, 2018`_.
+
+It is worth noting in either case that denoising to ASVs and clustering to OTUs are seperate, but parallel steps. A choice should be made for a single pathway: either denoising or OTU based clustering; it is not recommended to combine the steps.
+
+In this tutorial, we’ll denoise using Deblur on single ended sequences. those interested in Dada2 may find the :doc:`moving pictures tutorial  <moving-pictures/>` and :doc:`Atacama soil tutorial <atacama-soils>`. An example of using Deblur with paired end reads can be found in the :doc:`Alternative methods of read joining <read-joining/>` tutorial.
+
+Quality filtering
```
thermokarst: I am certainly not opposed to deblur here, but, one thing to keep in mind in a workshop environment is that the q2-deblur pipeline has a few more steps than the q2-dada2 pipeline. Food for thought. Also, we don't necessarily need to choose one vs the other in the doc itself, we can do something similar to the MP tutorial where we demonstrate _both_ pipelines in the doc, but only execute one pipeline in an instructional scenario.

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+   5. Which features are observed in at least 47 samples?
+
+Generating a Phylogenetic Tree for Diversity Analysis
+=====================================================
+
+QIIME 2 analysis allows the use of phylogenetic trees for both diversity metrics such as PD whole tree and UniFrac distance as well as feature-based analyses in Gneiss. The tree provides an inherent structure to the data, allowing us to consider an evolutionary relationship between organisms.
+
+QIIME 2 offers several ways to construct a phylogenetic tree. For this tutorial, we’re going to use a fragment insertion tree using the ``fragment-insertion`` plugin. The authors of the fragment insertion plugin suggest that it can outperform traditional alignment based methods based on short illumina reads by alignment against a reference tree built out of larger sequences. Our command, ``qiime fragment-insertion sepp`` will take the representative sequences (a ``FeatureData[Sequence]`` object) we generated during deblurring and return a phylogenetic tree where the sequences have been inserted into the greengenes 13_8 99% identity reference tree backbone.
```
thermokarst:
```suggestion
QIIME 2 offers several ways to construct a phylogenetic tree. For this tutorial, we’re going to use a fragment insertion tree using the ``fragment-insertion`` plugin. The authors of the fragment insertion plugin suggest that it can outperform traditional alignment based methods based on short Illumina reads by alignment against a reference tree built out of larger sequences. Our command, ``qiime fragment-insertion sepp`` will take the representative sequences (a ``FeatureData[Sequence]`` Artifact) we generated during deblurring and return a phylogenetic tree where the sequences have been inserted into the Greengenes 13_8 99% identity reference tree backbone.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+
+*Note: This command tables about 11 minutes to run.*
```
thermokarst:
```suggestion
.. note:: This command takes approximately 10 minutes to run when ``-p-n-threads`` is set to ``1``. If your computation environment supports it, we suggest including an appropriately-set ``--p-n-threads`` parameter.*
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+
+.. command-block::
+
+   qiime fragment-insertion sepp \
```
thermokarst: Let's include ``--p-n-threads 1 \`` in here to go with the note above.

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+   In how many samples were there reads exceeding the maximum number of ambiguous bases?
+
+Denoising
+---------
+
+Next, we’ll apply the Deblur algorithm with the ``qiime deblur denoise-16S`` command.
+
+The method requires the use of an additional parameter: ``p-trim-length``. This controls the length of the sequences and should be selected based on a drop in quality scores. In our dataset, the quality scores are relatively evenly distributed along the sequencing run, so we’ll use the full 150 bp sequences. However, the selection of the trim length is a relatively subjective measurement and relies on the decision making capacity of the analyst.
+
+*Note*: The command is expected to take about 3-4 minutes to run.
```
thermokarst:
```suggestion
.. note:: The command is expected to take about 3-4 minutes to run.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+   -  Unweighted UniFrac distance
+   -  Weighted UniFrac distance
+
+There is a very good discussion of diversity metrics and their meanings in a `qiime forum by Stephanie Orchanian`_.
+
+This method wraps several other methods, and it’s worthwhile to note that the steps in ``qiime diversity core-metrics-phylogenetic`` can be executed independently.
+
+One important consideration for diversity calculations is the Rarefaction depth. Above, we used alpha rarefaction and the sample summary to pick a rarefaction depth. So, for these analyses, we’ll use a depth of 1000 sequences per sample.
+
+*Note*: This step takes about 7 minutes
```
thermokarst:
```suggestion
.. note:: This step takes about 7 minutes
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+   qiime fragment-insertion sepp \
+     --i-representative-sequences ./deblur_rep_set.qza \
+     --o-tree ./tree.qza \
+     --o-placements ./tree_placements.qza
+
+
+Taxonomic Classification
+========================
+
+Let’s do one more preparation step before we dig into the analysis! To be able to identify ASVs and give them “names”, we need to somehow determine taxonomy. To do this, we’ll use the ``q2-feature-classifier`` plugin.
```
thermokarst:
```suggestion
Let’s do one more preparation step before we dig into the analysis! To be able to identify ASVs and give them "names", we need to somehow determine taxonomy. To do this, we’ll use the ``q2-feature-classifier`` plugin.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+
+For this analysis, we'll use a pretrained classifier using 99% Greengenes 13_8 reference set trimmed to 250 bp of the V4 hypervariable region (corresponding to the 515F-806R primers). The classifier is a specific semantic type, ``TaxonomicClassifier``, and it is actually the object that does the classification.
+
+.. download::
+   :url: https://data.qiime2.org/2019.4/common/gg-13-8-99-515-806-nb-classifier.qza
+   :saveas: gg-13-8-99-515-806-nb-classifier.qza
+
+It’s worth noting that naive bayesian classifiers perform best when they’re trained for the specific hypervariable region amplified. You can train a classifier specific for your dataset based on the :doc:`training classifiers tutorial <feature-classifier>` or download classifiers for other datasets from the :doc:`QIIME 2 resource page <../data-resources>`. Classifiers can be re-used for consistent versions of the underlying packages, database and region of interest.
```
thermokarst:
```suggestion
It’s worth noting that Naive Bayes classifiers perform best when they’re trained for the specific hypervariable region amplified. You can train a classifier specific for your dataset based on the :doc:`training classifiers tutorial <feature-classifier>` or download classifiers for other datasets from the :doc:`QIIME 2 resource page <../data-resources>`. Classifiers can be re-used for consistent versions of the underlying packages, database and region of interest.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+
+.. command-block::
+
+   qiime feature-classifier classify-sklearn \
+     --i-reads ./deblur_rep_set.qza \
+     --i-classifier ./gg-13-8-99-515-806-nb-classifier.qza \
+     --o-classification ./taxonomy.qza
+
+.. do we want to throw clawback in here?
```
thermokarst: @nbokulich - thoughts?

nbokulich: _maybe_... for the purposes of the workshop it would be neat to showcase clawback, but for the purposes of a general-use tutorial clawback may be a bit complicated (multiple additional analysis steps, etc).

thermokarst: I vote for skipping ATM, let's focus on Copenhagen as our initial target.

source/tutorials/pd-mice.rst
jwdebelius: I am so proud of myself that this is the first major spelling mistake at line 122

source/tutorials/pd-mice.rst
jwdebelius: Im happy either way

source/tutorials/pd-mice.rst
jwdebelius: One of the features I struggle with often in my own analysis, since I get handed enough datasets that are numerically coded.

source/tutorials/pd-mice.rst
jwdebelius: I think my experience with Qiita is one reason for seperation. But I'm not wed to the idea... just means updating the mapping file and data dictionary

source/tutorials/pd-mice.rst
jwdebelius: Semi person preference, semi what I want my students to use when I make them go through this tutorial because I prefer Deblur to Dada2, but it doesn't really matter to me here.

@nbokulich thoughts?

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,748 @@
...
+   2. Which metric shows saturation and stabilization of the diversity?
+   3.  Which mouse genetic background has higher diversity, based on the curve? Which has shallower sequencing depth?
+
+   Now, let's check the feature table summary.
+
+   1. What percentage of samples are lost if we set the rarefaction depth to 1250 sequences per sample?
+   2. Which mice did the missing samples come from?
+
+**Based on the current rarefaction curve and sample summary, what sequencing depth would you pick? Why?**
+
```
jwdebelius: I really want a shiny box here that draws attention to the discussion. But, IDK

source/tutorials/pd-mice.rst
nbokulich: deblur is faster, which is a bonus for workshops (though I don't know how that stacks up with the extra QC step)

Otherwise I agree with @thermokarst the fewer steps the better for a workshop...

jwdebelius: It looks like in this case, Dada2 is quicker than Deblur. So, I'm going to update the tutorial to use Dada2 for the workshop. Might take me a day or two and will slightly shift questions but shouldnt change conclusions.

source/tutorials/pd-mice.rst
nbokulich: Instead of using boldface text, use `.. note::` or `.. question::`. See the moving-pics tutorial for some examples of what this looks like [in the rst file](https://raw.githubusercontent.com/qiime2/docs/master/source/tutorials/moving-pictures.rst) vs. how it appears in the rendered doc... e.g., see [this section](https://docs.qiime2.org/2019.4/tutorials/moving-pictures/#alpha-rarefaction-plotting) for an example of both.

jwdebelius: Ive got those a couple other places. 

I want a pink box or something that highlights this. It's right below a question, so that doesn't stand out, and I tend to ignore notes, because, well, they're blue and thats probably not important. Which may just mean there are too many boxes

nbokulich: `.. warning::` renders a yellow box. `.. danger::` is red. See examples [here](https://docs.qiime2.org/2019.4/data-resources/). The only problem is that this note is neither a danger nor a warning.

jwdebelius: It's not a note, question, warning, or danger. It's a checkpoint or a decision point. 

How hard would it be to add a new pink box for checkpoint?

nbokulich: See [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html) for all possible directives. I am really not sure what would be involved in defining a new directive.

thermokarst: Besides adding a new directive, it is also possible to insert vanilla HTML into the doc (see [here](https://github.com/qiime2/docs/blame/master/source/install/native.rst#L27-L72) for an example).

You could do something like:

```sphinx
.. raw:: html
   <div class="admonition" style="background:#FF78CB;">
     <strong>
       Based on the current rarefaction curve and sample summary, what
       sequencing depth would you pick? Why?
     </strong>
   </div> 
```

which would render like:

image

:woman_singer: :man_singer:

jwdebelius: I just made a new directive, but the vanilla HTML might have been better. Pro either way is that now I know more about Sphinx directives?

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+.. command-block::
+
+   qiime feature-classifier classify-sklearn \
+     --i-reads ./deblur_rep_set.qza \
+     --i-classifier ./gg-13-8-99-515-806-nb-classifier.qza \
+     --o-classification ./taxonomy.qza
+
+.. do we want to throw clawback in here?
+
+Now, let’s review the taxonomy associated with the sequences using the ``qiime metadata tabulate`` function.
```
thermokarst:
```suggestion
Now, let’s review the taxonomy associated with the sequences using the ``qiime metadata tabulate`` method.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+
+.. command-block::
+
+   qiime metadata tabulate \
+     --m-input-file ./taxonomy.qza \
+     --o-visualization ./taxonomy.qzv
+
+Let’s also tabulate the representative sequences. Tabulating the representative sequences will allow us to see the sequence assigned to the identifier and interactively blast the sequence against the NCBI database.
```
thermokarst:
```suggestion
Let’s also tabulate the representative sequences (``FeatureData[Sequence]``). Tabulating the representative sequences will allow us to see the sequence assigned to the identifier and interactively blast the sequence against the NCBI database.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+
+   How many sequences are mapped to g__Akkermansia?
+
+   Use the tabulated representative sequences to look up these features. If you blast them against NCBI, do you get the same taxonomic identifier?
+
+
+Alpha Rarefaction and Selecting a Rarefaction Depth
+===================================================
+
+Although sequencing depth in a microbiome samples does not directly relate to the original biomass in a community, the relative sequencing depth has a large impact on observed communities (`Weiss et al, 2017`_). Therefore, for most diversity metrics, a normalization approach is needed.
```
thermokarst:
```suggestion
Although sequencing depth in a microbiome sample does not directly relate to the original biomass in a community, the relative sequencing depth has a large impact on observed communities (`Weiss et al, 2017`_). Therefore, for most diversity metrics, a normalization approach is needed.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+
+Current best practices suggest the use of rarefaction, a normalizational via sub sampling without replacement. Rarefaction occurs in two steps. First, samples which are below the rarefaction depth are filtered out of the feature table. Then, all remaining samples are subsampled without replacement to get to the sequencing depth. It’s both important and sometimes challenging to select a rarefaction depth for diversity analyses. Several strategies exist to figure out the right rarefaction depth, but alpha rarefaction is a data-driven way to approach the problem.
```
thermokarst:
```suggestion
Current best practices suggest the use of rarefaction, a normalizational via sub-sampling without replacement. Rarefaction occurs in two steps: first, samples which are below the rarefaction depth are filtered out of the feature table. Then, all remaining samples are subsampled without replacement to get to the sequencing depth. It’s both important and sometimes challenging to select a rarefaction depth for diversity analyses. Several strategies exist to figure out an appropriate rarefaction depth - we will primarily consider alpha rarefaction in this tutorial, because it is a data-driven way to approach the problem.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+   qiime diversity alpha-rarefaction \
+     --i-table ./deblur_table.qza \
+     --m-metadata-file ./metadata.tsv \
+     --o-visualization ./alpha_rarefaction_curves.qzv \
+     --p-min-depth 10 \
+     --p-max-depth 2500
+
+The visualization file will give us two curves. The top curve will give the alpha diversity (observed OTUs or shannon) as a function of the sequencing depth. This is used to determine whether the richness or evenness has saturated based on the sequencing depth. The rarefaction curve should “level out” as you approach a sequencing depth. Failure to do so, especially with a diversity-only metric such as observed OTUs or Faith’s PD diversity, may indicate that the richness in the samples has not been fully saturated.
+
+The second curve shows the number of samples in each group at each sequencing depth. This is useful to determine the sampling depth where samples are lost, and whether this may be biased by metadata group. Remember that rarefaction is a two step process and samples which do not meet the rarefaction depth are filtered out of the table. So, we can use the curves to look at the number of samples by different metadata categories.
```
thermokarst:
```suggestion
The second curve shows the number of samples in each group at each sequencing depth. This is useful to determine the sampling depth where samples are lost, and whether this may be biased by metadata column group values. Remember that rarefaction is a two step process and samples which do not meet the rarefaction depth are filtered out of the table. So, we can use the curves to look at the number of samples by different metadata columns.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+
+If you’re still unsure whether the rarefaction depth, you can also use the sample summary to look at which samples are lost by adding metadata to the feature table summary.
```
thermokarst:
```suggestion
If you’re still unsure of the rarefaction depth, you can also use the sample summary to look at which samples are lost by supplying sample metadata to the feature table summary.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+
+*Hint*: We generated this in the after we built the feature table.
```
thermokarst:
```suggestion
*Hint*: We generated this after we built the feature table.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+   -  Jaccard distance
+   -  Bray Curtis distance
+   -  Unweighted UniFrac distance
+   -  Weighted UniFrac distance
+
+There is a very good discussion of diversity metrics and their meanings in a `qiime forum by Stephanie Orchanian`_.
+
+This method wraps several other methods, and it’s worthwhile to note that the steps in ``qiime diversity core-metrics-phylogenetic`` can be executed independently.
+
+One important consideration for diversity calculations is the Rarefaction depth. Above, we used alpha rarefaction and the sample summary to pick a rarefaction depth. So, for these analyses, we’ll use a depth of 1000 sequences per sample.
```
thermokarst:
```suggestion
One important consideration for diversity calculations is the rarefaction depth. Above, we used the alpha rarefaction visualization and the sample summary visualization to pick a rarefaction depth. So, for these analyses, we’ll use a depth of 1000 sequences per sample.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+     --i-table ./deblur_table.qza \
+     --i-phylogeny ./tree.qza \
+     --m-metadata-file ./metadata.tsv \
+     --p-sampling-depth 1000 \
+     --output-dir ./core-metrics-results
+
+Alpha Diversity
+---------------
+
+Alpha diversity asks whether the number of distribution of features within a sample differ between different conditions. The comparison makes no assumptions about the features that are shared between samples; two samples can have the same alpha diversity and not share any features. The rarified alpha diversity produced by ``q2-diversity`` is a univariate, continuous value and can be tested using common non-parametric statistical tests.
```
thermokarst:
```suggestion
Alpha diversity asks whether the distribution of features within a sample differ between different conditions. The comparison makes no assumptions about the features that are shared between samples; two samples can have the same alpha diversity and not share any features. The rarified alpha diversity produced by ``q2-diversity`` is a univariate, continuous value and can be tested using common non-parametric statistical tests.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+   For this exercise, we'll look at the group significance results for Faith's phylogenetic diversity and evenness.
+
+   Based on the group significance test, is there a difference in phylogenetic diversity by genotype? Is there a difference based on the donor?
+
+If we had a continuous covariate that we thought was associated with the alpha diversity, we could test that using ``qiime diversity alpha-correlation``. However, the only continuous variable in this dataset is the days since transplant.
+
+Beta Diversity
+--------------
+
+Next, we’ll compare the structure of the microbiome communities using beta diversity. Start by making a visualize inspection of the principle coordinates plots (PCoA) plots that were generated by emperor and ``core-metrics-results/weighted_unifrac_emperor.qzv`` into `view.qiime2.org`_
```
thermokarst:
```suggestion
Next, we’ll compare the structure of the microbiome communities using beta diversity. Start by making a visual inspection of the principle coordinates plots (PCoA) plots that were generated by emperor and ``core-metrics-results/weighted_unifrac_emperor.qzv``.
```

Just as a general note, q2view is not the only way to view results, I would prefer if we keep the "viewing" language agnostic, just to prevent confusion.

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+
+.. command-block::
+
+   qiime diversity adonis \
+     --i-distance-matrix core-metrics-results/unweighted_unifrac_distance_matrix.qza \
+     --m-metadata-file metadata.tsv \
+     --o-visualization core-metrics-results/unweighted_adonis.qzv \
+     --p-formula genotype+donor
+
+.. do we also want permadisp here?
```
thermokarst: I will defer to you and @nbokulich here.

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+
+.. command-block::
+
+   qiime feature-table filter-features \
+     --i-table ./table_1k.qza \
+     --p-min-frequency 50 \
+     --p-min-samples 4 \
+     --o-filtered-table ./table_1k_abund.qza
+
+ANCOM operates on a ``FeatureTable[Composition]`` Artifact, which is based on the relative abundance of features on a per-sample basis. However, the ``FeatureTable[Composition]`` object cannot tolerate zeros (because compositional methods typically use a log-transform or a ratio and you can’t take the log or divide by zeros). To remove the zeros from our table, we add a pseudocount to the ``FeatureTable[Frequency]`` object.
```
thermokarst:
```suggestion
ANCOM fundamentally operates on a ``FeatureTable[Frequency]`` which is based on the frequencies of features on a per-sample basis. However, ANCOM cannot tolerate zeros (because compositional methods typically use a log-transform or a ratio and you can’t take the log or divide by zeros). To remove the zeros from our table, we add a pseudocount to the ``FeatureTable[Frequency]`` object, creating a ``FeatureTable[Composition]`` in its place.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+     --m-metadata-column donor \
+     --o-visualization ./ancom_donor.qzv
+
+   qiime composition ancom \
+     --i-table ./table1k_abund_comp.qza \
+     --m-metadata-file ./metadata.tsv \
+     --m-metadata-column genotype \
+     --o-visualization ./ancom_genotype.qzv
+
+When you open the ancom visualizations, you’ll see a volcano plot on top which relates the ANCOM W statistical to the CLR (center log transform) for the groups. The W statistic is the number of tests whether the ratio between a given pair of ASVs is significant at the test threshold (typically fdr-adjusted p < 0.05). Because differential abundance in ANCOM is based on the ratio between tests, it does not produce a traditional p-value.
```
thermokarst:
```suggestion
When you open the ANCOM visualizations, you’ll see a volcano plot on top which relates the ANCOM W statistical to the CLR (center log transform) for the groups. The W statistic is the number of tests whether the ratio between a given pair of ASVs is significant at the test threshold (typically FDR-adjusted p < 0.05). Because differential abundance in ANCOM is based on the ratio between tests, it does not produce a traditional p-value.
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+
+.. command-block::
+
+   qiime emperor plot \
+     --i-pcoa ./core-metrics-results/unweighted_unifrac_pcoa_results.qza \
+     --m-metadata-file ./metadata.tsv \
+     --p-custom-axes days_post_transplant \
+     --o-visualization ./core-metrics-results/unweighted_unifrac_emperor_time_axis.qzv
+
+.. included to recapitulate the step from moving pictures. Im not sure it adds much, and it covered there?
```
thermokarst: I don't think this is particularly useful, and usually isn't that interesting in workshops, I would be fine with dropping the custom axis example.

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+     --p-state-column days_post_transplant \
+     --p-individual-id-column mouse_id \
+     --p-group-columns genotype,donor \
+     --o-visualization ./from_first_unifrac_lme.qzv
+
+Now, let’s look at the results of the models.
+
+.. question::
+
+   Open the distance volatility plot (``./from_first_unifrac_vol.qzv``) using the qiime 2 viewer. Based on the volatility plot, does one donor change more over time than the other? What about by genotype? Cage?
```
thermokarst:
```suggestion
   Based on the volatility plot, does one donor change more over time than the other? What about by genotype? Cage?
```

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+
+   Now, let’s open the linear mixed effects model (``./from_first_unifrac_lme.qzv``). Is there a significant association between the genotype and temporal change? Which genotype is more stable (has lower variation)? Is there a temporal change associated with the donor? Did you expect or not expect this based on the volatility plot results? Can you find an interaction between the donor and genotype?
```
thermokarst:
```suggestion
   Now, let’s open the linear mixed effects model. Is there a significant association between the genotype and temporal change? Which genotype is more stable (has lower variation)? Is there a temporal change associated with the donor? Did you expect or not expect this based on the volatility plot results? Can you find an interaction between the donor and genotype?
```

source/tutorials/pd-mice.rst
thermokarst: Should this be moved up in the tutorial? Like, before we run the LME action?

source/tutorials/pd-mice.rst
```diff
@@ -0,0 +1,749 @@
...
+
+But, we saw differences by donor and genotype based on beta diversity. Using the PCoA, we can see clear separation between the mice from the two donors (this recapitulates the results of the original paper). After adjusting for the donor, we saw a significant difference between the genotypes.
+
+Although there wasn’t a clear pattern in the barchart at the phylum level between donors or genotypes, we were still able to find ASVs which differentiated the genotypes at using ANCOM. There was no overlap between these ASVs in the donor and genetic background, supporting the hypothesis that the difference due to genotype is seperate from the difference due to donor.
+
+The volatility plots and temporal analysis showed the microbiome in different genetic backgrounds changed differently over time.
+
+This suggests that there is an effect on the microbiome of mice receiving fecal transplants due to genotype.
+
+.. Next steps?
```
thermokarst: :party_parrot:

jwdebelius:
> Lookin' great! Finished the second half of my pass-through. Next up, I am going to run commands and attempt to answer your Q&A sections.

That might be dangerous. I haven't quite finished the Dada2 run through. So, you'll be the first! Hopefully, nothing major changes!

source/tutorials/pd-mice.rst
jwdebelius: is "column group values" the official QIIME language?

source/tutorials/pd-mice.rst
jwdebelius: Cool. I need to look into some of those.
And oh God, sorry about the proof reading stuff.

source/tutorials/pd-mice.rst
jwdebelius: Good. Let's drop it, then. I think volatility plots are better and moving pictures covers it if people need it.

source/tutorials/pd-mice.rst
jwdebelius: Conceptually, I like combining visualisations with statistical models because I think visualisations can help give you the intuition for the statistics. So, it may be better to have it with the volitility plots int he workshop setting, and may be something to keep in a different order for teaching

source/theme/static/style.css
```diff
@@ -64,7 +64,14 @@ div.admonition.warning {
 div.admonition.question {
     color: #3C763D;
     background-color: #DFF0D8;
-    border: 1px solid #D6E9C6;
+    border: 1px solid #FFB7D2;
```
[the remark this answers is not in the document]
jwdebelius: 😦 I think I fixed it in the latest push. Crap. Def not intended!

source/tutorials/pd-mice.rst
jwdebelius: I like it from a conceptual perspective and practice running the command, but Im not sure it adds that much. @nbokulich?

nbokulich: I think a brief mention/example would be nice. We do not talk about permdisp in other tutorials; unless if we write an in-depth tutorial on diversity analyses at some point (not a bad idea!), it could be good to start using permdisp in our workshop tutorial(s).

jwdebelius: Okay... I added permdisp for cage. Its still not a great example (I sort of wish I had overlap in this tutorial), but that may be for a later, more indepth option. Id be interested in contributing to that sort of thing at a later date.

PR #407, opened 2019-05-15
MAINT: No more QIIME 1-esque MD columns

### LINKED ISSUE #396: "Modernize" metadata headers in tutorial metadata files
**Improvement Description**
"Modernize" metadata headers in tutorial metadata files.

**Current Behavior**
Most or all of the current tutorials use the old QIIME 1 "`#SampleID`" header label.

**Proposed Behavior**
Update the sample ID header label to a new, non-Q1 label.

**Questions**
1. Which ID label? I was thinking `sample-id`...
2. Should we also standardize the other column labels?

thermokarst:
Fixes #396

### DIFF
--- a/source/tutorials/atacama-soils.rst
+++ b/source/tutorials/atacama-soils.rst
@@ -84,13 +84,13 @@ To analyze these data, the sequences that you just downloaded must first be impo
       --input-path emp-paired-end-sequences \
       --output-path emp-paired-end-sequences.qza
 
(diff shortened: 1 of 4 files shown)


thermokarst:
Ran `make html` locally, all seems to work.

Not all tutorials were updated (for example, q2-gneiss), but the "main" tutorials. 

Also, square bracket column quoting is the new "standard" for sql-where clauses!

PR #413, opened 2019-06-05
text tweaks plus sample classifier

nbokulich:
@thermokarst @jwdebelius 

1. fixed some typos and some unclear sentences.
2. note: the use of first-distances previously looked at rate of change, not difference from day 7 (as was indicated in the text). I added the `baseline` parameter so that the command conforms to the text — but if you prefer to look at rate of change pls revert that command and modify the text instead.
3. I moved the taxonomy classification section until after diversity. This matches what we will do in the workshop, and makes a bit more sense (phylogeny happens next to diversity, and taxonomy happens next to barcharts).
4. I added a short q2-sample-classifier demo to the end of the tutorial. This is my new favorite tutorial dataset for sample classifiers, though it is no good for regression (so we may still need to use the ECAM dataset in the workshop for that purpose).

### DIFF
--- a/source/tutorials/pd-mice.rst
+++ b/source/tutorials/pd-mice.rst
@@ -692,8 +692,8 @@ Let's use the ``q2-longitudinal`` plugin to look at how samples from an individu
      --m-metadata-file ./core-metrics-results/unweighted_unifrac_pcoa_results.qza \
      --p-state-column days_post_transplant \
      --p-individual-id-column mouse_id \
-     --p-default-group-column 'donor_status'\
-     --p-default-metric 'Axis 2'\
+     --p-default-group-column 'donor_status' \
+     --p-default-metric 'Axis 2' \
      --o-visualization ./pc_vol.qzv
 
 .. question::
@@ -729,7 +729,7 @@ We can again use volatility analysis to visualize the change in beta diversity b
      --p-state-column days_post_transplant \
      --p-individual-id-column mouse_id \
      --p-default-metric Distance \
-     --p-default-group-column 'donor_status'\
+     --p-default-group-column 'donor_status' \
      --o-visualization ./from_first_unifrac_vol.qzv
 
 .. question::
--- a/source/tutorials/pd-mice.rst
+++ b/source/tutorials/pd-mice.rst
@@ -1,9 +1,9 @@
 Parkinson's Mouse Tutorial
 --------------------------
 
-This tutorial will demonstrate a "typical" QIIME 2 analysis, using a set of fecal samples from `humanized`_ mice. The original study, `Sampson et al, 2016`_, was designed to determine whether the fecal microbiome contributed to the development of Parkinson's Disease (PD). Several observation studies showed a difference in the microbiome between PD patients and controls, although the organisms identified across studies were not consistent. However, this was sufficient evidence to suggest that there might be a relationship between PD and the fecal microbiome.
+This tutorial will demonstrate a "typical" QIIME 2 analysis of 16S rRNA gene amplicon data, using a set of fecal samples from `humanized`_ mice. The original study, `Sampson et al, 2016`_, was designed to determine whether the fecal microbiome contributed to the development of Parkinson's Disease (PD). Several observation studies showed a difference in the microbiome between PD patients and controls, although the organisms identified across studies were not consistent. However, this was sufficient evidence to suggest that there might be a relationship between PD and the fecal microbiome.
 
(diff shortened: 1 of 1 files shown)


source/tutorials/pd-mice.rst
```diff
@@ -432,7 +380,7 @@ One important consideration for diversity calculations is the rarefaction depth.
 Alpha diversity
 +++++++++++++++
 
-Alpha diversity asks whether the distribution of features within a sample differ between different conditions. The comparison makes no assumptions about the features that are shared between samples; two samples can have the same alpha diversity and not share any features. The rarefied alpha diversity produced by ``q2-diversity`` is a univariate, continuous value and can be tested using common non-parametric statistical tests.
+Alpha diversity asks whether the distribution of features within a sample (or groups of samples) differs between different conditions. The comparison makes no assumptions about the features that are shared between samples; two samples can have the same alpha diversity and not share any features. The rarefied alpha diversity produced by ``q2-diversity`` is a univariate, continuous value and can be tested using common non-parametric statistical tests.
```
jwdebelius:
```suggestion
Alpha diversity asks whether the distribution of features within a sample (or groups of samples) differ between different conditions. The comparison makes no assumptions about the features that are shared between samples; two samples can have the same alpha diversity and not share any features. The rarefied alpha diversity produced by ``q2-diversity`` is a univariate, continuous value and can be tested using common non-parametric statistical tests.
```

source/tutorials/pd-mice.rst
```diff
@@ -460,6 +408,17 @@ We can test our covariates of interest against Faith's phylogenetic diversity an
...
+In some experiments, multiple interacting factors may impact alpha diversity together. If our alpha diversity estimates follow a normal distribution, we may use analysis of variance (ANOVA) to test whether multiple effects significantly impact alpha diversity. This test is present in the ``q2-longitudinal`` plugin:
+
+.. command-block::
+
+   qiime longitudinal anova \
+     --m-metadata-file ./core-metrics-results/faith_pd_vector.qza \
+     --m-metadata-file ./metadata.tsv \
+     --p-formula 'faith_pd ~ genotype * donor_status' \
+     --o-visualization ./core-metrics-results/faiths_pd_anova.qzv
+
```
jwdebelius: That's shiny. Although I never would have thought to look for it in q2-longitudinal. Good for the richness metrics which usually are asymptotically normal

source/tutorials/pd-mice.rst
nbokulich: the distribution ... differs
not
the distribution ... differ

source/tutorials/pd-mice.rst
nbokulich: yeah @thermokarst and I had discussed back and forth about where to add it... I originally put it in q2-longitudinal thinking that it could be useful for simple longitudinal analyses and because it has many of the same dependencies and underlying functions that the LME visualizer uses. It just sort of stuck. Right now it is definitely hidden so having it in a tutorial like this would help make it more visible.

source/tutorials/pd-mice.rst
```diff
@@ -749,22 +770,72 @@ Now, let's look at the results of the models.
...
 Synthesis
 =========
 
 Based on the results of the analysis, we can say that there is a difference in the microbial communities of these mice based on their donor and genetic background. (This recapitulates the results of the original analysis.)
 
 We found that the donor is the primary driver of alpha diversity.
 
-But, we saw differences by donor and genotype based on beta diversity. Using the PCoA, we can see clear separation between the mice from the two donors (this recapitulates the results of the original paper). After adjusting for the donor, we saw a significant difference between the genotypes.
+But, we saw differences by donor and genotype based on beta diversity. Using the PCoA emperor plots, we can see clear separation between the mice from the two donors (this recapitulates the results of the original paper). After adjusting for the donor, we saw a significant difference between the genotypes.
```
jwdebelius: I know this was my text, but...

```suggestion
But, we saw differences by donor and genotype based on beta diversity. Using the PCoA emperor plots, we can see clear separation between the mice from the two donors. After adjusting for the donor, we saw a significant difference between the genotypes.
```

nbokulich: LGTM thanks!

source/tutorials/pd-mice.rst
```diff
@@ -749,22 +770,72 @@ Now, let's look at the results of the models.
...
 
-Although there wasn't a clear pattern in the barchart at the phylum level between donors or genotypes, we were still able to find ASVs which differentiated the genotypes at using ANCOM. There was no overlap between these ASVs in the donor and genetic background, supporting the hypothesis that the difference due to genotype is separate from the difference due to donor.
+The volatility plots and temporal analysis showed that the microbiome in different genetic backgrounds changed differently over time.
 
-The volatility plots and temporal analysis showed the microbiome in different genetic backgrounds changed differently over time.
+This suggests that there is a genotype-specific effect on the microbiome of mice receiving fecal transplants.
 
-This suggests that there is an effect on the microbiome of mice receiving fecal transplants due to genotype.
+💩🐁
```
jwdebelius: 👍

source/tutorials/pd-mice.rst
nbokulich: actually one really neat thing that could go in that visualizer if you are interested is a test for normality.

jwdebelius: This gets back to the ongoing discussion about needing a diversity-specific tutorial that keeps coming up

nbokulich: Definitely! One day...

source/tutorials/pd-mice.rst
jwdebelius: Having verified, it still sounds weird, but eh. Grammar has never been my best subject

thermokarst:
Going to merge this PR as-is - @jwdebelius, if you have additional suggestions or edits we can push them up in another round of changes. Thanks so much you two!

PR #415, opened 2019-06-06
IMP: Python-based command blocks

thermokarst:
@ebolyen, can you take a look at this when you get a chance? I'm not quite sure why I am not able to catch stdout from my python blocks, maybe I just missed something simple. Thanks!

### DIFF
--- a/source/sphinx_extensions/command_block/extension.py
+++ b/source/sphinx_extensions/command_block/extension.py
@@ -14,6 +14,7 @@ import subprocess
 import tempfile
 import urllib.parse
 import functools
+import sys
 
 import docutils.nodes
 import docutils.parsers.rst
@@ -65,6 +66,7 @@ class CommandBlockDirective(docutils.parsers.rst.Directive):
     has_content = True
 
     option_spec = {
+        'runtime': docutils.parsers.rst.directives.unchanged_required,
         'no-exec': docutils.parsers.rst.directives.flag,
         'url': docutils.parsers.rst.directives.unchanged_required,
         'saveas': docutils.parsers.rst.directives.unchanged_required,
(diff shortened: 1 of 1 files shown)


source/sphinx_extensions/command_block/extension.py
```diff
@@ -128,13 +139,26 @@ def run(self):
...
+            'shell': self._execute_sh_commands,
+            'python': self._execute_py_commands,
+        }[runtime](commands, working_dir)
+
+    def _execute_py_commands(self, commands, working_dir):
+        with tempfile.NamedTemporaryFile(mode='w', encoding='utf-8') as fh:
+            for command in commands:
+                fh.write('%s\n' % command)
+            return self._execute_sh_commands(
+                ['%s %s' % (sys.executable, str(fh.name))], working_dir)
```
ebolyen: Is there ever a need to share scope across blocks for this?

thermokarst: Probably, and that is certainly cumbersome w this current approach (would require Artifacts to be saved to disk).

ebolyen:
When you say stdout isn't working, is it like `print` isn't working? I don't see how it would fail...

ebolyen:
Alternatively, would a doctest directive work instead? We used that in the dev docs and it generally works pretty well.

ebolyen:
In particular: https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testcode might do what you need.

thermokarst:
Yeah, none of my print statements will work. Will look into the `testcode` directive, that might get me there.

thermokarst:
Okay, that directive doesn't quite do it for me. I wanted this functionality for the Artifact API section of my WIP "Utilities" tutorial, but I think instead I will just skip that interface (for now). Maybe @Oddant1 wants to take over this PR?

PR #416, opened 2019-06-07
MAINT: command-block - handle result overwriting

thermokarst:
This code was originally written by @ebolyen


ebolyen:
I've never seen this code in my life, officer ... I plead the fifth ... it was under duress

PR #418, opened 2019-06-07
IMP: Utility functions in QIIME 2 Tutorial

thermokarst:


source/tutorials/utilities.rst
```diff
@@ -0,0 +1,151 @@
...
+
+.. command-block::
+   :stdout:
+
+   qiime tools --help
+
+
+Let's get our hands on some data so that we can learn more about this
+functionality! First, we will take a look at the taxonomic bar charts from the
+Moving Pictures Tutorial<LINK>:
```
nbokulich: maybe we should use artifacts from the PD tutorial instead?

nbokulich: I don't have strong opinions btw, and expect you have good reason for using MP

source/tutorials/utilities.rst
```diff
@@ -0,0 +1,151 @@
...
+Now that we know how to use the command, we will run the following:
+
+.. command-block::
+   :stdout:
+
+   qiime tools citations mp-taxa-bp.qzv
+
+As you can see, the citations for this particular visualization are presented
+above in bibtex format.
+
```
nbokulich: maybe also show how to get citations for a plugin? I realize it is a little out of scope for the utilities tutorialbut a very brief mention would be useful (or link to another resource — I think there's a forum post where I describe various ways of checking citations?)

source/tutorials/utilities.rst
```diff
@@ -0,0 +1,151 @@
...
+Viewing Visualizations
+......................
+
+What if we want to view our bar plots? One option is to load the visualization
+at https://view.qiime2.org. Another option is to use ``qiime tools view`` to
+accomplish the job (please note, provenance viewing is only available at
```
nbokulich: make this note a note directive?

source/tutorials/utilities.rst
```diff
@@ -0,0 +1,151 @@
...
+Validating Results
+..................
+
+We can also validate the integrity of the file by running
+``qiime tools validate``:
+
+.. command-block::
+   :stdout:
+
+   qiime tools validate faith_pd_alpha_div.qza
```
nbokulich: maybe you should add an invalid file! to demonstrate that, e.g., fasta files can be imported but only minimal validation is performed on import

thermokarst: Let's add that in a follow-up PR.

thermokarst:
Screenshot_2020-02-14 Utilities in QIIME 2 — QIIME 2 2020 2 0 dev0+15 g73b431a documentation

source/tutorials/utilities.rst
```diff
@@ -0,0 +1,151 @@
...
+   qiime tools citations mp-taxa-bp.qzv
+
+As you can see, the citations for this particular visualization are presented
+above in bibtex format.
+
+Viewing Visualizations
+......................
+
+What if we want to view our bar plots? One option is to load the visualization
+at https://view.qiime2.org. Another option is to use ``qiime tools view`` to
```
nbokulich: are there any reasons to use `view` instead of view.qiime2.org? May be worth mentioning here. The one obvious one is that you don't need to be connected to the internet... though don't want to give the impression that the data "go" anywhere! This might be a good place to repeat the disclaimer that view does not transmit any data ever...

maybe this is just one of those comments that opens a can of worms and proves totally unhelpful 😞

source/tutorials/utilities.rst
nbokulich: I agree, not necessary in this PR, just something that might be useful for workshops and teaching.

source/tutorials/utilities.rst
```diff
@@ -0,0 +1,173 @@
...
+
+   qiime tools inspect-metadata sample-metadata.tsv
+
+.. question::
+
+   How many metadata columns are there in ``sample-metadata.tsv``? How many IDs?
+   Identify how many categorical columns are present. Now do the same for numerical
+   columns.
+
+This tool can be very helpful for learning about Metadata column names for
```
nbokulich: OMG inspect metadata to get column names! I learn/forget something new every day!

source/tutorials/utilities.rst
```diff
@@ -0,0 +1,173 @@
...
+.. question::
+
+   How many IDs are there? How many columns? Are there any categorical columns? Why?
+
+Artifact API
+------------
+
+.. TODO: finish this section
+
+Coming soon, please stay tuned!
```
nbokulich: can't wait!

source/tutorials/utilities.rst
thermokarst: Let's not open that can-o-worms...

PR #420, opened 2019-06-14
BUG: fix multi-command command-blocks

ebolyen:
With the addition of IO capturing, we lost the ability to run multiple commands in a single block. This broke a few tutorials which groups commands in a single logical block.

Fixing the early return and limiting IO capture to a single command per block seems to do the trick.


ebolyen:
pinging @thermokarst for perusal later next week (will merge before then)

thermokarst:
Thanks!

thermokarst:
Thanks @ebolyen & @ChrisKeefe. I have proposed changes to this in #422 ---just needed to iterate over the processes in the `_get_stream_node` method, no need to limit things to one process per block.

PR #422, opened 2019-06-18
IMP: Support multi-command streamed blocks

thermokarst:
image


thermokarst:
This reverts the changes introduced in #420 in order to support stream capturing in multi-command blocks.

PR #353, opened 2018-09-20
adding fmt tutorial

antgonza:
Brief summary of the Pull Request, including any issues it may fix using the GitHub closing syntax:

https://help.github.com/articles/closing-issues-using-keywords/

Also, include any co-authors or contributors using the GitHub coauthor tag:

https://help.github.com/articles/creating-a-commit-with-multiple-authors/

---

Include any questions for reviewers, screenshots, sample outputs, etc.

PR #424, opened 2019-06-21
Fmt cdiff khanna

antgonza:
This is the new FMT C. Diff tutorial using the Khanna dataset. This is based on the [previous version](https://github.com/qiime2/docs/pull/353) just adapting the text and commands to this new dataset. 

Note that it uses SEPP and that the commands are slower than the old version.

cc: @ElDeveloper, @thermokarst

### DIFF
--- a/bin/workshop.py
+++ b/bin/workshop.py
@@ -17,7 +17,7 @@ tutorial_data_url = ('https://docs.qiime2.org/%s/data/tutorials/%s' %
 tutorials_url = ('https://docs.qiime2.org/%s/tutorials/' %
                  (qiime2.__release__,))
 tutorial_url = ''.join([tutorials_url, tutorial_name])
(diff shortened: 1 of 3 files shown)


thermokarst:
@antgonza - where should I get the sample metadata from? This tutorial won't build until the MD is in place. Just send me a TSV if you have a new file, otherwise, please point me in the right direction.

cc @ElDeveloper

thermokarst:
Nevermind! I found it! Forgot we had a google drive folder with all this stuff in it. Stay tuned!

lizgehret:
Hey @gregcaporaso @ebolyen looks like we never ended up adding this tutorial - should I just close this PR?

gregcaporaso:
Yes, we can close this. IIRC, this was an early draft of what became the [_Current protocols_](https://currentprotocols.onlinelibrary.wiley.com/doi/10.1002/cpbi.100) paper, the content of which can also be found [here](https://curr-protoc-bioinformatics.qiime2.org/).

PR #445, opened 2020-02-06
Remove building linear models section

### LINKED ISSUE #444: Revise q2-gneiss tutorial to reflect removed visualizers
**Improvement Description**
https://github.com/qiime2/q2-gneiss/pull/67 removed all 4 visualizers found in the q2-gneiss plugin:

- `balance-taxonomy`
- `dendrogram-heatmap`
- `lme-regression`
- `ols-regression`

The [q2-gneiss tutorial](https://github.com/qiime2/docs/blob/master/source/tutorials/gneiss.rst) contains examples of three of these visualizers in action. We need to update the tutorial to not include these examples.

**Current Behavior**
Removed Visualizers are referenced in the tutorial.

**Proposed Behavior**
The Visualizer references should be removed from the tutorial.

**Questions**
1. @mortonjt - what should we do about the prose surrounding those commands? Is it acceptable to just remove the ["Building linear models" section](https://github.com/qiime2/docs/blob/master/source/tutorials/gneiss.rst#building-linear-models-using-balances)?

**References**
1. https://github.com/qiime2/q2-gneiss/pull/67

andrewsanchez:
Closes #444


mortonjt:
I think it is a little awkward to cut out half of the tutorial - it'll probably just make more sense to delete the whole thing.

mortonjt:
Thought about this a little more - the other methods (i.e. ilr hierarchical or ilr phylogenetic) are not deprecated.  Those methods are pretty solid and probably should be directly integrated with other differential abundance methods at some point.

For now, another option is to keep the heatmap visualization so that there is some sensible endpoint to this analysis.  The future of these ilr-based methods is really bottlenecked by phylogenetic visualizations - so until a scalable/customizable phylogenetic visualization is ready (i.e. empress), we won't be able to tie up these tutorials.

thermokarst:
Thanks @mortonjt! So to make sure I am understanding correctly, it sounds like maybe we should bring back the `dendrogram-heatmap` visualizer, that way we can keep most of this tutorial section, and have a reasonable endpoint?

mortonjt:
Right.  We can still have the `dendrogram-heatmap` deprecated, but at least keep it in the tutorial so that there is still an endpoint (otherwise it'll be hard to see why we'd need to have this tutorial).

thermokarst:
Thanks @mortonjt, that plan works for me, stay tuned!

@andrewsanchez - can you update the PR to bring back everything up to ~L166, delete everything after "These taxa in the denominator could be interesting".

Also, @andrewsanchez, can you cut a PR on q2-gneiss that adds the `dendrogram-heatmap` back in, while still keeping the `deprecated=True` flag? Basically undo part of https://github.com/qiime2/q2-gneiss/pull/67.

Thanks!

PR #456, opened 2020-03-19
ENH: Protocols paper updates

thermokarst:
Fixes #455

PR #462, opened 2020-05-23
Updates from source activate to conda activate

ChrisKeefe:
This PR updates this repo (content and travis commands) to use the contemporary `conda activate`

Conda 4.6 was [released](https://www.anaconda.com/blog/conda-4-6-release) in June 2019, so `conda activate` has been [preferred by conda](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#activating-an-environment) over `source activate` for nearly a year. This command unifies the conda activation/deactivation UI across all platforms, so may reduce friction for conda users native to Windows.


thermokarst:
Thanks @ChrisKeefe! This has been on our radar for a while, but we were waiting for the kinks to be worked out, I think now is a great time to roll it out. Two quick notes, I removed the bug labels, since this is not related to a bug, just improving the docs to go along with the state of the conda project (the `source` syntax is still supported by conda) . The second note is that native Windows installations of QIIME 2 still aren't supported, only WSL Windows installations, so dealing with native windows shells is a non-issue (WSL is a sort of lightweight Linux VM, so WSL users use Linux shells, like bash or zsh).

I'm sending this back to you for now, there are some build/test issues that still need to be sorted out. Let me know if you need a hand.

Thanks!

.travis.yml
```diff
@@ -13,7 +13,7 @@ before_install:
 install:
   - wget -q https://raw.githubusercontent.com/qiime2/environment-files/master/latest/staging/qiime2-latest-py36-linux-conda.yml
   - conda env create -q -n test-env --file qiime2-latest-py36-linux-conda.yml
-  - source activate test-env
+  - conda activate test-env
```
thermokarst, line 16: You'll need to run `conda init` to get the shell configured, first.

source/install/native.rst
```diff
@@ -99,7 +99,7 @@ Now that you have a QIIME 2 environment, activate it using the environment's nam
 
    conda activate qiime2-2020.5
 
-To deactivate an environment, run ``source deactivate``.
+To deactivate an environment, run ``conda deactivate``.
```
thermokarst, line 102: Good catch! I updated everything back in https://github.com/qiime2/docs/pull/456 to use `conda activate`, looks like I missed this line.

source/install/native.rst
```diff
@@ -153,4 +153,4 @@ command and look for the line that starts with "active environment":
    conda info
 
 .. _`Miniconda`: https://docs.conda.io/en/latest/miniconda.html
-.. _`Miniconda instructions`: https://conda.io/projects/conda/en/latest/user-guide/install/index.html
\ No newline at end of file
+.. _`Miniconda instructions`: https://conda.io/projects/conda/en/latest/user-guide/install/index.html
```
thermokarst, line 156: What changed on this line?

ChrisKeefe: Sorry about the noop. Looks like VSCode might have added a newline to the end of the file, and I missed it in review.

image

ChrisKeefe:
Sorry if my description was unclear, @thermokarst. Command parity across OS may help
> users native to Windows

even though they're working on Linux/MacOS, because they're being moved to a single, shared command `conda activate`, rather than to a unix-like only command `source activate`. When they get back to windows, they can conda without having to worry about multiple commands.

.travis.yml
ChrisKeefe: `conda init` will use bash by default, and Catalina uses `zsh` IIRC. When/if we add a darwin workflow to this sucker, is that something we should be dealing with? If so, I can open an issue.

ChrisKeefe:
Note: `conda init` fails to fix the breaking build. Instead, am going to try directly sourcing the required functions, per the [sample](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/use-conda-with-travis-ci.html) provided by conda.

.travis.yml
ChrisKeefe: Note: Based on the official docs (and a failed attempt to use `conda init` here), we probably won't have access to `conda init` and its defaults. if we want to handle multiple shells, we may have to explicitly source the shell-appropriate scripts. See [SO](https://stackoverflow.com/questions/55342122/conda-activate-on-travis-ci).

.travis.yml
```diff
@@ -6,14 +6,16 @@ before_install:
   - wget -q https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
   - export MINICONDA_PREFIX="$HOME/miniconda"
   - bash miniconda.sh -b -p $MINICONDA_PREFIX
+  - source "$HOME/miniconda/etc/profile.d/conda.sh"
+  - hash -r
```
thermokarst, line 10: Is this line necessary?

```bash
help hash
hash: hash [-lr] [-p pathname] [-dt] [name ...]
    For each NAME, the full pathname of the command is determined and
    remembered.  If the -p option is supplied, PATHNAME is used as the
    full pathname of NAME, and no path search is performed.  The -r
    option causes the shell to forget all remembered locations.  The -d
    option causes the shell to forget the remembered location of each NAME.
    If the -t option is supplied the full pathname to which each NAME
    corresponds is printed.  If multiple NAME arguments are supplied with
    -t, the NAME is printed before the hashed full pathname.  The -l option
    causes output to be displayed in a format that may be reused as input.
    If no arguments are given, information about remembered commands is displayed.
```

If it isn't, please remove (that help text above makes it sound like something we don't need, but, I'm not sure). If it is, can you just make a note of why it's needed, or how things fail if it isn't included?

source/install/native.rst
thermokarst: Looks like this still unresolved, can you please undo this part of the changeset?

.travis.yml
ChrisKeefe: That line clears the shell's pathname cache. It is probably not needed, but was recommended in the conda example, presumably because conda is doing interesting things with the shell's env variables. I've dropped it in my most recent PR, and it will stay dropped unless it breaks something. : )

PR #465, opened 2020-06-02
HACK: Add note about vega, revert later

Oddant1:
It looks like the only plot that mentions the vega editor in its doc text is `longitudinal volatility`. This PR adds a note about that feature being temporarily disabled pending a rework of this plot and others to work properly with Vega 5.

### DIFF
--- a/source/tutorials/longitudinal.rst
+++ b/source/tutorials/longitudinal.rst
@@ -131,7 +131,7 @@ In the resulting visualization, a line plot appears on the left-hand side of the
 6. Group mean lines and spaghetti can also be modified with the "scatter size" and "scatter opacity" slider bars in the "Plot Controls". These adjust the size and opacity of individual points. Maximize scatter opacity and minimize line opacity to transform these into longitudinal scatter plots!
 7. Relevant sample metadata at individual points can be viewed by hovering the mouse over a point of interest.
 
-If the interactive features of this visualization don't quite scratch your itch, click on the "Open in Vega Editor" button at the top of the "Plot Controls" and customize to your heart's content. This opens a window for manually editing plot characteristics in `Vega Editor`_, a visualization tool external to QIIME2.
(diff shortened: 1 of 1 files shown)


thermokarst:
We could also tweak the existing text to say something like "you can copy the spec into the vega editor (URL here) and edit away"

nbokulich:
yeah I like that idea... though we should add instructions on accessing the vega spec in that case

Oddant1:
Sorry, I was looking into something else. Is the decision to tell them where to go externally to use the editor and how to get the spec to use with the editor? Isn't the editor still not going to work correctly if they do that?

thermokarst:
> Is the decision to tell them where to go externally to use the editor and how to get the spec to use with the editor?

Yes, I think so - does that sound good to you?

> Isn't the editor still not going to work correctly if they do that?

Yeah, not without edits - I think we can write the text in a way that doesn't commit us one way or the other - let us know if you want some suggestions or need a hand.

Thanks!

source/tutorials/longitudinal.rst
```diff
@@ -131,7 +131,7 @@ In the resulting visualization, a line plot appears on the left-hand side of the
 6. Group mean lines and spaghetti can also be modified with the "scatter size" and "scatter opacity" slider bars in the "Plot Controls". These adjust the size and opacity of individual points. Maximize scatter opacity and minimize line opacity to transform these into longitudinal scatter plots!
 7. Relevant sample metadata at individual points can be viewed by hovering the mouse over a point of interest.
 
-If the interactive features of this visualization don't quite scratch your itch, click on the "Open in Vega Editor" button at the top of the "Plot Controls" and customize to your heart's content. This opens a window for manually editing plot characteristics in `Vega Editor`_, a visualization tool external to QIIME2 (NOTE: This feature is temporarily disabled. Apologies for the inconvenience) .
+If the interactive features of this visualization don't quite scratch your itch, you can use the external vega editor to edit the plot. Click on the "View Source" button then copy the contents of that page and paste them into the left column of this webpage https://vega.github.io/editor/#/edited.
```
thermokarst, line 134: Can you trim the editor url to just `https://vega.github.io/editor/`? Thanks!

PR #467, opened 2020-06-15
MAINT: changes observed_otus to observed_features in tutorials

ChrisKeefe:
Updates the Moving Pictures and Parkinson's Mouse tutorials, changing `observed_otus` to `observed_features` to align with `q2-diversity-lib`


ChrisKeefe:
Screencap of built note: 
image

source/tutorials/moving-pictures.rst
```diff
@@ -255,6 +255,8 @@ QIIME 2's diversity analyses are available through the ``q2-diversity`` plugin,
   * unweighted UniFrac distance (a qualitative measure of community dissimilarity that incorporates phylogenetic relationships between the features)
   * weighted UniFrac distance (a quantitative measure of community dissimilarity that incorporates phylogenetic relationships between the features)
 
+.. note:: 🏗👷 Some descriptions are changing in QIIME 2's ``diversity`` tools. The phrase ``observed otus`` is being replaced with ``observed features``, because "features" better describes the different ways in which users work with taxonomic features. This will affect both documentation and (in places) the names of command arguments. For a short time, you will see both phrases, but these measures of diversity are identical `under the hood <http://scikit-bio.org/docs/latest/generated/skbio.diversity.alpha.observed_otus.html#skbio.diversity.alpha.observed_otus>`_.
```
thermokarst, line 258: Great! I have some ideas for edits, below.

- I think we can drop the inline-code formatting for the phrases "observed otus" and "observed features" - I think emphasis by wrapping normal quotes around will be sufficient, but if you still want to emphasize more, you could make it boldface or italicized.
- RE "better describes the different ways in which users work with taxonomic features" - should that actually say _non-taxonomic features_, instead? The main reason for renaming from "OTUs" to "features" is to decouple the metric from the typical amplicon sequencing study. Features can be things like OTUs or ASVs (or taxons), but they can also be things like metabolites, or pathways.
- Please drop the "For a short time" phrase - we can discuss timing in the changelog. As well, this release of QIIME 2 will always have this discrepancy, so the "short time" phrase (without additional explanation) is potentially misleading.

ChrisKeefe:
@thermokarst , do we want to drop this same note into the "same" location in the PD mouse tutorial?
pd_mice

thermokarst:
> @thermokarst , do we want to drop this same note into the "same" location in the PD mouse tutorial?

Sure! That's a great idea!

PR #486, opened 2020-09-04
TYPO: grammar

ChrisKeefe:

PR #488, opened 2020-09-14
PD-Mice Tutorial Edits from CZI Workshop Sprints

Oddant1:


thermokarst:
As discussed here (https://github.com/qiime2/docs/pull/486#issuecomment-689291262), let's keep this PR open long enough to collect _everyone's_ various PD Mice changes.

andrewsanchez:
@ChrisKeefe, would you mind pushing your commits in Oddant1/docs/pull/1 to the branch associated with this PR instead so that we can integrate the changes in one review/commit?  I think that was what @thermokarst was requesting [here](https://github.com/qiime2/docs/pull/489#issuecomment-695136879).  Let me know if you need a hand with that .  Thanks 🙏 !

andrewsanchez:
I noticed a possible discrepancy between PD-mice and Moving Pictures regarding selection max depth for rarefaction.  Moving Pictures states:

>In general, choosing a value that is somewhere around the median frequency seems to work well, but you may want to increase that value if the lines in the resulting rarefaction plot don’t appear to be leveling out, or decrease that value if you seem to be losing many of your samples due to low total frequencies closer to the minimum sampling depth than the maximum sampling depth.

While PD-mice states:

> We want to set a maximum depth close to the maximum number of sequences.

If this is something worth addressing in this PR, should we expound a bit in PD-mice to agree with what Moving Pictures says?

thermokarst:
I don't think this is a discrepancy - the MP note is providing some general guidance, while the PD Mice is provide _dataset specific_ recommendations. I think the PD Mice text (in context) makes that mission fairly clear (emphasis mine, to show that the line you plucked out was sandwiched in between a lot of important information/discussion):

> We’ll use qiime diversity alpha-rarefaction to subsample the ASV table at different depths (between --p-min-depth and --p-max-depth) and calculate the alpha diversity using one or more metrics (--p-metrics). When we checked the feature table, we found that the sample with the fewest sequences in the denoised table has 85 features and that the sample with the most has 4996 features. **We want to set a maximum depth close to the maximum number of sequences.** We also know that if we look at a sequencing depth around 4250 sequences per sample, we’ll be looking at information from 22 samples. So, let’s set this as our maximum sequencing depth.

source/tutorials/pd-mice.rst
```diff
@@ -293,7 +293,7 @@ We now have a feature table (observation matrix) of ASVs in each sample and a ph
...
+Rarefaction is a normalization via sub-sampling without replacement and is commonly used as a workaround for this issue. Rarefaction occurs in two steps: first, samples which are below the rarefaction depth are filtered out of the feature table. Then, all remaining samples are subsampled without replacement to get to the specified sequencing depth. It's both important and sometimes challenging to select a rarefaction depth for diversity analyses. Several strategies exist to figure out an appropriate rarefaction depth - we will primarily consider alpha rarefaction in this tutorial, because it is a data-driven way to approach the problem.
```
thermokarst, line 296:
```suggestion
Rarefaction is a method for normalization via sub-sampling without replacement and is commonly used as a workaround for the issue of uneven sequencing depth. Rarefaction occurs in two steps: first, samples which are below the rarefaction depth are filtered out of the feature table. Then, all remaining samples are subsampled without replacement to get to the specified sequencing depth. It's both important and sometimes challenging to select a rarefaction depth for diversity analyses. Several strategies exist to figure out an appropriate rarefaction depth - we will primarily consider alpha rarefaction in this tutorial, because it is a data-driven way to approach the problem.
```

andrewsanchez (approved):
Thanks, @ChrisKeefe and @Oddant1!  🤓

PR #498, opened 2021-01-26
Install gevent and nodejs with conda.

### LINKED ISSUE #497: Newer versions of pip are problematic when installing q2studio
**Bug Description**
The QIIME 2 Studio interface docs (Ref 1 below) show users how to install dependencies using pip. It looks like this is now failing on many user's machines (Refs 2 and 3), presumably because of newer versions of pip. It would be great to replace this with @andrewsanchez's instructions (Ref 3) that leverage conda:

```bash
conda install gevent nodejs
```

**References**
1. https://github.com/qiime2/docs/blob/master/source/interfaces/q2studio.rst
2. https://forum.qiime2.org/t/error-when-installing-qstudio-2020-11-0/18226
3. https://forum.qiime2.org/t/unable-to-run-q2studio/17920

andrewsanchez:
See #497 for details.  closes #497

### DIFF
--- a/source/interfaces/q2studio.rst
+++ b/source/interfaces/q2studio.rst
@@ -36,6 +36,7 @@ Next we need to install it (both as a Python package, and as a Node.js package):
 .. command-block::
    :no-exec:
 
+   conda install gevent nodejs
    pip install .
    npm install && npm run build
 
--- a/source/interfaces/q2studio.rst
+++ b/source/interfaces/q2studio.rst
@@ -36,7 +36,7 @@ Next we need to install it (both as a Python package, and as a Node.js package):
 .. command-block::
    :no-exec:
 
-   conda install gevent nodejs
+   conda install gevent nodejs -c defaults --override-channels
    pip install .
    npm install && npm run build
 


source/interfaces/q2studio.rst
```diff
@@ -36,6 +36,7 @@ Next we need to install it (both as a Python package, and as a Node.js package):
 .. command-block::
    :no-exec:
 
+   conda install gevent nodejs
```
thermokarst, line 39: This should probably have some explicit channel(s) listed (`-c defaults ...` etc), as well as `--override-channels` to ensure that a user's local conda config doesn't install the wrong thing.

See here for an example of this, in action:

https://github.com/qiime2/busywork/blob/644376876df798ef34f29c0494be596f9335f36e/ci/master/bin/test.sh#L10-L14

thermokarst: I should say, the main reason I propose adding explicit channels in (and disabling the channels in their local conda config) is mostly just a support thing - it helps us ensure that they are installing the same versions of things (from the same sources), as us. Functionally its probably no different to omit the explicit channels.

andrewsanchez: Sounds good to me!

thermokarst (approved):

PR #516, opened 2021-07-02
update references re: rescript

nbokulich:
Just some little tweaks to the data-resources page:

* added citation for the RESCRIPt pre-print (currently cites the zenodo doi linked to the GH repo).

* Also slightly adjusted the description and references for the weighted classifiers to indicate (a) that they are not only V4 but they are 16S, and (b) to also cite the references listed for the unweighted classifiers.


lizgehret (approved):
lgtm, thanks @nbokulich!

PR #517, opened 2021-07-15
MAINT: doc updates for q2cli cast-metadata tool

lizgehret:
Docs added for q2cli cast-metadata tool. PR for this tool can be found [here](https://github.com/qiime2/q2cli/pull/261) (still in draft PR format).


source/tutorials/metadata.rst
```diff
@@ -122,9 +122,15 @@ Column Types
...
 
-You can use an optional *comment directive* to declare column types in your metadata file. The comment directive must appear **directly** below the header. The row's first cell must be ``#q2:types`` to indicate the row is a *comment directive*. Subsequent cells may contain the values ``categorical`` or ``numeric`` (both case-insensitive). The empty cell is also supported if you do not wish to assign a type to a column (the type will be inferred in that case). Thus, it is easy to include this comment directive without having to declare types for every column in your metadata.
+.. note:: Columns which contain **categorical** metadata (e.g. non-numeric data) cannot be set to numeric column types, due to the way that metadata column types are parsed. This is because integers can be set to string values, but strings cannot be set to integer values.
```
thermokarst, line 127: Thanks, this'll be a good note to include! One issue though - not all numeric data is comprised of integers - there is also support for floating-point numbers (floats). Perhaps you can just make this note more explicit to solve the problem "Columns which contain _non-numerical_ metadata (e.g. **categorical** data) cannot be set to the `numeric` column type.

source/tutorials/metadata.rst
```diff
@@ -122,9 +122,15 @@ Column Types
...
+
+For manual specifications within your metadata file(s), the comment directive must appear **directly** below the header. The row's first cell must be ``#q2:types`` to indicate the row is a *comment directive*. Subsequent cells may contain the values ``categorical`` or ``numeric`` (both case-insensitive). The empty cell is also supported if you do not wish to assign a type to a column (the type will be inferred in that case). Thus, it is easy to include this comment directive without having to declare types for every column in your metadata.
+
+This functionality is now also supported directly through QIIME 2's command line interface (as of the 2021.8 release), via the q2cli cast-metadata developer tool. This tool allows for bulk specifications to your metadata file(s) column types, set to either **categorical** or **numerical**. This tool functions exactly the same as the aformentioned comment directive, but allows for inline data manipulation (or the ability to automate column type assignment through a custom script), which can be a more robust method than manual file manipulation.
```
thermokarst, line 133: - No need to mention the 2021.8 release directly - the page will say it in multiple places, and it might be a point of confusion post-2021.8 (folks might think it is _only_ in 2021.8).
- "q2cli cast-metadata developer tool" -> `qiime tools cast-metadata` tool. It isn't really a "developer tool" so I suggest dropping that. You could also include a link to the q2cli interface page within the User Docs.
- Can you show an example or two of how to use this command? That's the best part about these docs is that there are interactive tutorials!

thermokarst (changes requested):

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,82 @@ run, using ``inspect-metadata``, we can learn all about it:
...
+In this example, we will cast the ``days_post_transplant`` column from ``numeric`` to ``categorical``,
+and the ``mouse_id`` column from ``categorical`` to ``numeric``. The rest of the columns contained
+within our metadata will be left as-is.
+
+.. command-block::
+   :no-exec:
+
+   qiime tools cast-metadata sample_metadata.tsv \
+   --cast days_post_transplant:categorical --cast mouse_id:numeric \
+   --output-file casted_metadata.tsv
```
thermokarst, line 205: If you drop the `:no-exec:` then this command will actually run. If you put the `:stdout:` flag on and drop `--output-file` then we'll get a nice rendered output:

```rst
.. command-block::
   :stdout:

   qiime tools cast-metadata sample_metadata.tsv \
   --cast days_post_transplant:categorical \
   --cast mouse_id:numeric
```

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,82 @@ run, using ``inspect-metadata``, we can learn all about it:
...
+The ``--ignore-extra`` and ``--error-on-missing`` flags are used to handle casted columns not contained within the original
+metadata file, and columns contained within the metadata file that aren't included in the cast call, respectively.
+We can take a look at how these flags can be used below:
+
+In the first example, we'll take a look at utilizing the ``--ignore-extra`` flag when a column is cast that is not included
+within the original metadata file. Let's start by looking at what will happen if an extra column is included and this flag
+is not enabled.
+
+.. command-block::
+   :stdout:
```
thermokarst, line 225: I think the flag here should be `:stderr:`, not `:stdout:`

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,82 @@ run, using ``inspect-metadata``, we can learn all about it:
...
+   :allow-error:
+
+   qiime tools cast-metadata sample_metadata.tsv \
+   --cast spleen:numeric
+
+Notice that the ``spleen`` column included in the cast call results in a raised error. If we want to ignore any extra columns that
+are not present in the original metadata file, we can enable the ``--ignore-extra`` flag.
+
+.. command-block::
+   :no-exec:
```
thermokarst, line 235:
```suggestion
   :stdout:
```

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,82 @@ run, using ``inspect-metadata``, we can learn all about it:
...
+
+In our second example, we'll take a look at the ``--error-on-missing`` flag, which handles columns that are present within the
+metadata that are not included in the cast call.
+
+The default behavior permits a subset of the full metadata file to be included in the cast call (e.g. not all columns within
+the metadata must be present in the cast call). If the ``--error-on-missing`` flag is enabled, all metadata columns must be
+included in the cast call, otherwise an error will be raised.
+
+.. command-block::
+   :stdout:
```
thermokarst, line 250:
```suggestion
   :stderr:
```

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,82 @@ run, using ``inspect-metadata``, we can learn all about it:
...
+Below we will go through a few scenarios of how this tool can be used, and some common mistakes
+that may come up.
+
+We'll start by first downloading some sample metadata. **Note**: This is the same sample
+metadata used in the **Inspect Metadata** section, so you can skip this step if you have
+already downloaded the ``sample_metadata.tsv`` file from above.
+
+.. download::
+   :url: https://data.qiime2.org/2021.8/tutorials/pd-mice/sample_metadata.tsv
+   :saveas: sample-metadata.tsv
```
thermokarst, line 194:
```suggestion
   :saveas: sample_metadata.tsv
```

thermokarst (commented):
```
make preview DEBUG=tutorials/utilities && cd build/preview && python -m http.server ; cd - 
```

source/tutorials/metadata.rst
```diff
@@ -122,9 +122,15 @@ Column Types
...
+
+For manual specifications within your metadata file(s), the comment directive must appear **directly** below the header. The row's first cell must be ``#q2:types`` to indicate the row is a *comment directive*. Subsequent cells may contain the values ``categorical`` or ``numeric`` (both case-insensitive). The empty cell is also supported if you do not wish to assign a type to a column (the type will be inferred in that case). Thus, it is easy to include this comment directive without having to declare types for every column in your metadata.
+
+This functionality is now also supported directly through QIIME 2's command line interface, via the ``qiime tools cast-metadata`` tool. This tool allows for bulk specifications to your metadata file(s) column types, set to either **categorical** or **numerical**. This tool functions exactly the same as the aformentioned comment directive, but allows for inline data manipulation (or the ability to automate column type assignment through a custom script), which can be a more robust method than manual file manipulation. A detailed tutorial and example usage for this tool can be found on the `QIIME 2 Utilities`_ tutorial page.
```
thermokarst, line 133: - `s/numerical/numeric/g`
- "This tool functions exactly the same as..." - the `cast-metadata` tool _is_ using the comment directive - as it is written now it kinda sounds like there is some alternative mechanism for declaring column types, which isn't the case. Can you tweak the language here to clear that up?

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,78 @@ run, using ``inspect-metadata``, we can learn all about it:
 
    How many IDs are there? How many columns? Are there any categorical columns? Why?
 
+Casting Metadata
```
thermokarst, line 180: Maybe "Casting Metadata Columns" ? Or "Casting Metadata Column Types" ?

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,78 @@ run, using ``inspect-metadata``, we can learn all about it:
...
+
+In this example, we will cast the ``days_post_transplant`` column from ``numeric`` to ``categorical``,
+and the ``mouse_id`` column from ``categorical`` to ``numeric``. The rest of the columns contained
+within our metadata will be left as-is.
+
+.. command-block::
+   :stdout:
+
+   qiime tools cast-metadata sample_metadata.tsv \
+   --cast days_post_transplant:categorical --cast mouse_id:numeric
```
thermokarst, line 204: Can you split this into two separate lines, that will match the general style of these docs, and in general I have found it also helps make the content of a command a bit more clear.

```bash
--cast days_post_transplant:categorical \
--cast mouse_id:numeric
```

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,78 @@ run, using ``inspect-metadata``, we can learn all about it:
...
+
+If the ``--output-file`` flag is enabled, the specified output file will contain the modified column types that we cast above
+for the original ``sample_metadata.tsv``.
```
thermokarst, line 207: I'm not sure I understand what this "for the original" means - can you explain, or revise the language here?

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,78 @@ run, using ``inspect-metadata``, we can learn all about it:
...
+
+If you do not wish to save your casted metadata to an output file, you can omit the ``--output-file`` parameter and
```
thermokarst, line 209: `s/casted/cast/g`

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,78 @@ run, using ``inspect-metadata``, we can learn all about it:
...
+   --cast spleen:numeric
+
+Notice that the ``spleen`` column included in the cast call results in a raised error. If we want to ignore any extra columns that
+are not present in the original metadata file, we can enable the ``--ignore-extra`` flag.
+
+.. command-block::
+   :stdout:
+
+   qiime tools cast-metadata sample_metadata.tsv \
+   --cast spleen:numeric --ignore-extra
```
thermokarst, line 234: Same comment as above RE splitting flags up one per line

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,78 @@ run, using ``inspect-metadata``, we can learn all about it:
...
+   :allow-error:
+
+   qiime tools cast-metadata sample_metadata.tsv \
+   --cast spleen:numeric
+
+Notice that the ``spleen`` column included in the cast call results in a raised error. If we want to ignore any extra columns that
+are not present in the original metadata file, we can enable the ``--ignore-extra`` flag.
+
+.. command-block::
+   :stdout:
```
thermokarst, line 231: On second thought, maybe we should drop the `:stdout:` here and just let folks know that we are omitting the output --- it is long and doesn't really serve much purpose being printed here a second time.

source/tutorials/utilities.rst
```diff
@@ -177,6 +177,78 @@ run, using ``inspect-metadata``, we can learn all about it:
...
+The default behavior permits a subset of the full metadata file to be included in the cast call (e.g. not all columns within
+the metadata must be present in the cast call). If the ``--error-on-missing`` flag is enabled, all metadata columns must be
+included in the cast call, otherwise an error will be raised.
+
+.. command-block::
+   :stderr:
+   :allow-error:
+   
+   qiime tools cast-metadata sample_metadata.tsv \
+   --cast mouse_id:numeric --error-on-missing
```
thermokarst, line 250: Same comment as above RE splitting flags up one per line

thermokarst (changes requested):
This is great, and more-or-less ready to go, just a few minor tweaks. Content-wise, this is just what we needed.

Two other requests for change here:

- Please remove all trailing whitespace
- Please wrap lines to ~95 columns

source/tutorials/utilities.rst
lizgehret: @thermokarst can you clarify what your requested change is for this code block? I'm not sure I understand what `s/casted/cast/g` is meant to replace.

thermokarst: Sure - my bad, we chatted about that syntax a while ago, I assumed you remembered it - it means "substitute the first value with the second value, everywhere" - its really common syntax, you'll see it in lots of places like sed, vim, your favorite editor, etc.

thermokarst: Sorry, I should clarify - its commonly-adopted syntax that you'll see pop up in lots of places, so it'll be good to file this one away because I bet you'll see it many more times!

lizgehret: Ahhh, okay gotcha! I do remember seeing this in vim, but I didn't realize this was common shorthand outside of that. Thanks!

thermokarst (approved):
Awesome, thanks so much @lizgehret & @Oddant1! These additions will be really helpful! I have a few very minor formatting tweaks, inline. I will commit those tweaks, then once CI has passed I will merge this PR. Thanks!

PR #525, opened 2021-11-21
replacing conda env creation with mamba for speed up

valentynbez:
Brief summary of the Pull Request, including any issues it may fix using the GitHub closing syntax:
- Added a line in installation section to replace `conda` with `mamba` for environment creation to speed up the process

## Tested on QIIME2 2020.8
```
time conda env create -n qiime2-2020.8 --file qiime2-2020.8-py36-linux-conda.yml 
real    5m23,541s
user    3m42,669s
sys     0m20,728s
```

After env removal and cleaning cache with `conda clean -a`

```
time mamba env create -n qiime2-2020.8 --file qiime2-2020.8-py36-linux-conda.yml 
real    3m15,572s
user    2m19,527s
sys     0m20,032s
```


valentynbez:
In release `2022.8` dependency resolution with `conda` takes a huge amount of time, so installing plugins becomes a problem.
`q2-krona` is the most lightweight plugin with a single dependency.
```bash
time conda install -c bioconda krona

real    1m52.516s
user    1m9.069s
sys     0m39.924s
```

```bash
time mamba install -c bioconda krona

real    0m13.355s
user    0m22.896s
sys     0m17.989s
```
```

**Comment by lizgehret** [comment]
This is a great suggestion @valentynbez, thank you! Looks like we have some merge conflicts due to changes within our docs from the time that this PR was originally opened. I'll get those updated and see how things are looking after that.

**Comment by lizgehret** [comment]
I think we'll also need to decide how we want to include mamba in our instructions here - whether we want to add it as a line before `mamba env create ...etc` or if we want to just add instructions in our conda install section to utilize mamba. cc @ebolyen

**Comment by valentynbez** [comment]
@lizgehret I think it would be useful to point users to `mamba` adding smth like:

Due to a lot of libraries needed for QIIME2 to properly function `conda` might be slow. In that case use `mamba`.  
```
conda install -y -c conda-forge mamba
mamba env create....
```

lizgehret:
Hey @valentynbez, thanks again for this suggestion! We are going to implement this change into the current release docs, but have added it in as a separate PR along with a few other installation changes/updates. I'm going to go ahead and close this, but we will make sure to include that this was your suggestion in our release changelog. Thanks!

lizgehret:
Hey @valentynbez upon testing the installation with the latest version of conda, it appears that conda and mamba are incompatible within the same environment. However, conda is now utilizing the libmamba solver (currently usable with the `--solver` parameter set to `libmamba`) and this will be set as the default solver in conda's next software update. So we've decided to revert our install instructions back to using `conda env create`, as this will be the most consistent instruction for users that prevents adding any dependencies to the environment that may cause conflicts (as is the case with the current version of conda).

Just wanted to give you a heads up since we removed this update from the release changelog and our user documentation. But thanks again for the suggestion, and we're looking forward to the latest conda update that will utilize libmamba - it will drastically cut down on install time for future releases!

PR #532, opened 2022-05-18
Update moving-pictures.rst

brett-van-tussler:
Typo fix, qualitiative to qualitative

---

Include any questions for reviewers, screenshots, sample outputs, etc.

PR #533, opened 2022-05-18
Update moving-pictures-usage.rst

brett-van-tussler:
Typo fix, qualitiative to qualitative

---

Include any questions for reviewers, screenshots, sample outputs, etc.


thermokarst:
@brett-van-tussler - do you think you could include this change in #532? That way everything is bundled together.

PR #537, opened 2022-08-10
drop link to sina community tutorial

nbokulich:
@nick-youngblut reported on [the forum](https://forum.qiime2.org/t/q2-alignment-reference-based-alignment-using-sina/6220/3?u=nicholas_bokulich) that SINA is no longer supported in q2-alignment, but a community tutorial for this deprecated action is still described in the QIIME 2 documentation.

This PR fixes this issue by:
1. removing the link to the now-archived tutorial 
2. instead ports over the relevant text describing the advantages of reference-based alignment
3. briefly describes how reference-based alignments can be imported into QIIME 2

Thanks @nick-youngblut for reporting this documentation lapse!

### DIFF
--- a/source/tutorials/phylogeny.rst
+++ b/source/tutorials/phylogeny.rst
@@ -119,14 +119,15 @@ plugin.
 
 *Reference based alignments*
 
-There are a variety of tools such as `PyNAST`_) (using `NAST`_), `Infernal`_,
-and `SINA`_, etc., that attempt to reduce the amount of ambiguously aligned
-regions by using curated reference alignments (e.g. `SILVA`_. Reference
+There are several tools that attempt to reduce the amount of ambiguously aligned
+regions by using curated reference alignments. Traditional, de novo alignment methods mututally align a set of unaligned sequences to create a multiple sequence alignment (MSA) from scratch. Re-running these methods with additional sequences will create MSAs with varying numbers of columns and assignments of bases to each column. These alignments is therefore incompatible with one another and may not be joined through concatenation.
+
+Reference based alignments, on the other hand, are meant to add sequences to an existing alignment. Alignments computed using reference based alignment tools always have widths identical to the reference alignment and maintain the meaning of each column. Therefore, these alignments may be concatenated.
+
+QIIME 2 currently does not wrap any methods for reference-based alignments, but alignments created using these methods can be imported into QIIME 2 as ``FeatureData[AlignedSequence]`` artifacts, provided that the alignments are standard FASTA formats. Some examples of tools for reference-based alignment include `PyNAST`_ (using `NAST`_), `Infernal`_, and `SINA`_. `SILVA`_ Reference
 alignments are particularly powerful for rRNA gene sequence data, as knowledge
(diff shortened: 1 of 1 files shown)


source/tutorials/phylogeny.rst
```diff
@@ -119,14 +119,15 @@ plugin.
 
 *Reference based alignments*
 
-There are a variety of tools such as `PyNAST`_) (using `NAST`_), `Infernal`_,
-and `SINA`_, etc., that attempt to reduce the amount of ambiguously aligned
-regions by using curated reference alignments (e.g. `SILVA`_. Reference
+There are several tools that attempt to reduce the amount of ambiguously aligned
+regions by using curated reference alignments. Traditional, de novo alignment methods mututally align a set of unaligned sequences to create a multiple sequence alignment (MSA) from scratch. Re-running these methods with additional sequences will create MSAs with varying numbers of columns and assignments of bases to each column. These alignments is therefore incompatible with one another and may not be joined through concatenation.
+
+Reference based alignments, on the other hand, are meant to add sequences to an existing alignment. Alignments computed using reference based alignment tools always have widths identical to the reference alignment and maintain the meaning of each column. Therefore, these alignments may be concatenated.
```
ebolyen, line 125: Is this always true? What if a query sequence has a unique and large insertion relative to the reference? I would expect the existing alignment to change to compensate. The individual coordinates could still be relative to the reference, but the number of columns would be larger.

source/tutorials/phylogeny.rst
```diff
@@ -119,14 +119,15 @@ plugin.
...
+
+QIIME 2 currently does not wrap any methods for reference-based alignments, but alignments created using these methods can be imported into QIIME 2 as ``FeatureData[AlignedSequence]`` artifacts, provided that the alignments are standard FASTA formats. Some examples of tools for reference-based alignment include `PyNAST`_ (using `NAST`_), `Infernal`_, and `SINA`_. `SILVA`_ Reference
 alignments are particularly powerful for rRNA gene sequence data, as knowledge
 of secondary structure is incorporated into the curation process, thus
-increasing alignment quality. For a more in-depth and eloquent overview of
-reference-based alignment approaches, check out the great `SINA community
-tutorial`_).
```
ebolyen, line 129: We'll need to delete the corresponding link at the bottom of the RST file.

ebolyen (commented):
Thanks @nbokulich!

source/tutorials/phylogeny.rst
nbokulich: I ported this text over verbatim from the community tutorial written by @epruesse so presumably this is the case at least for SINA, but I suppose this behavior might be tool-specific. 

We could also just drop this info altogether if it does not hold for other methods. @epruesse any comment on this?

source/tutorials/phylogeny.rst
nbokulich: 🤦 thanks! I missed that... fixed now

ebolyen:
I'm going to merge as is. It's still better than before, and we can update the text on reference alignments later.

Thanks @nbokulich!

PR #540, opened 2022-08-31
maint: removing vbox/aws details on install pages

lizgehret:
As of the 2022.8 release, we've decided to no longer build aws and vbox images for each release. To prevent any confusion moving forward, we are removing all associated install instructions from our docs.


ebolyen (approved):
LGTM, thanks @lizgehret!

PR #541, opened 2022-09-13
IMP: updating conda installation code block to separate install and cleanup commands

lizgehret:
The code blocks mixed in with text in our install instructions are a bit confusing and can lead to accidental copying of the text in addition to just the install commands. I made some small tweaks to the HTML block that I think will improve readability.

Here is what the install code block per OS **currently** looks like for conda installation:

**OSX**:
Screen Shot 2022-09-13 at 11 24 41 AM
**WSL**:
Screen Shot 2022-09-13 at 11 24 50 AM

And here is what these code blocks will look like (tested and rendered locally) when the text is separated from the actual commands:

**OSX**:
Screen Shot 2022-09-13 at 11 22 11 AM
**WSL**:
Screen Shot 2022-09-13 at 11 22 19 AM

I couldn't find a way to include the copy widget in this block due to the fact that this isn't just a raw code block - it's HTML, which accommodates the multiple panes per OS. If anyone knows of a way to add in the copy widget, please feel free to add that in!

### DIFF
--- a/source/install/native.rst
+++ b/source/install/native.rst
@@ -66,19 +66,19 @@ like for the environment. In this example, we'll name the environment
               From the above tabs, please choose the installation instructions that are appropriate for your platform.
             </p>
          </div>
-         <div id="macOS" class="tab-pane highlight-shell">
+         <div id="macOS" class="tab-pane fade">
             <pre>wget https://data.qiime2.org/distro/core/qiime2-2022.11-py38-osx-conda.yml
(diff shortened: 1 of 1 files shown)


Oddant1:
Looks good to me. I like having the comments separated out like that because it will (probably) get people to actually read and understand what's going on a bit better.

Oddant1 (approved):

PR #544, opened 2022-10-06
imp: adding M1 chip macOS install instructions

lizgehret:
This PR adds an install tab with instructions for macOS users with an M1 chip.

### DIFF
--- a/source/install/native.rst
+++ b/source/install/native.rst
@@ -78,6 +78,8 @@ like for the environment. In this example, we'll name the environment
    CONDA_SUBDIR=osx-64 conda env create -n qiime2-2022.11 --file qiime2-2022.11-py38-osx-conda.yml
    conda activate qiime2-2022.11
    conda config --env --set subdir osx-64</pre>
+   OPTIONAL CLEANUP
+   <pre>rm qiime2-2022.11-py38-osx-conda.yml</pre>
          </div>
          <div id="linux" class="tab-pane fade">
             <pre>wget https://data.qiime2.org/distro/core/qiime2-2022.11-py38-linux-conda.yml
--- a/source/install/native.rst
+++ b/source/install/native.rst
@@ -76,7 +76,7 @@ like for the environment. In this example, we'll name the environment
          <div id="M1-macOS" class="tab-pane fade">
             <pre>wget https://data.qiime2.org/distro/core/qiime2-2022.11-py38-osx-conda.yml
    CONDA_SUBDIR=osx-64 conda env create -n qiime2-2022.11 --file qiime2-2022.11-py38-osx-conda.yml
-   conda activate qiime2-2022.8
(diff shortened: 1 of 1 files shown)


lizgehret:
Preview of local docs build:
Screen Shot 2022-10-06 at 10 23 48 AM

Note that the `conda activate qiime2-2022.8` line has been updated to include the dev cycle epoch (2022.11). This screenshot was taken before that commit was pushed up (commit SHA [f30dab0](https://github.com/qiime2/docs/pull/544/commits/f30dab025ba80099f16110923f20b93162da7262)).

Screen Shot 2022-10-06 at 10 24 03 AM

ebolyen:
Any chance we can limit our tab length at this point so we can fit more in one line?

- OS X (64-bit)
- OS X (M1)
- Linux (64-bit)
- Windows Subsystem for Linux (64-bit)

cherman2 (approved):
LGTM!

source/install/native.rst
```diff
@@ -71,6 +72,12 @@ like for the environment. In this example, we'll name the environment
    conda env create -n qiime2-2022.11 --file qiime2-2022.11-py38-osx-conda.yml</pre>
    OPTIONAL CLEANUP
    <pre>rm qiime2-2022.11-py38-osx-conda.yml</pre>
+         </div>
+         <div id="M1-macOS" class="tab-pane fade">
+            <pre>wget https://data.qiime2.org/distro/core/qiime2-2022.11-py38-osx-conda.yml
+   CONDA_SUBDIR=osx-64 conda env create -n qiime2-2022.11 --file qiime2-2022.11-py38-osx-conda.yml
+   conda activate qiime2-2022.11
+   conda config --env --set subdir osx-64</pre>
```
ebolyen, line 80: Should we have the usual optional cleanup block after this?

lizgehret: oh whoops, yes that's a good idea. adding now!

lizgehret:
Update with clean-up block:

Screen Shot 2022-10-12 at 1 56 31 PM

ebolyen (approved):
Thanks @lizgehret!

thermokarst:
Food for thought - the "M1" instructions should just be the OS X instructions, period. Maybe people don't need to know which processor type they have, and the two commands are functionally identical. Also, there are other apple silicon processors besides "M1" - making this a bit more confusing for those folks.

https://en.wikipedia.org/wiki/Apple_silicon#M_series

thermokarst:
This feels like a premature optimization to me, but either way, M2 apple silicon is out now, so calling these the "M1" instructions might not be particularly clarifying either, FWIW.

thermokarst:
Relevant PR for the M2 discussion here: https://github.com/qiime2/docs/pull/545

PR #546, opened 2022-11-21
rm: removing q2studio from docs pages

lizgehret:
this PR removes all references to q2studio in our user documentation.


Oddant1:
The test failure looks like it's happening in sphinx?

Oddant1:
I don't understand what's happening here. In sphinx, they call importlib_metadata.entry_points with two kwargs, group and name, but the function takes no args or kwargs at all

PR #547, opened 2023-01-04
Greengenes database updates

wasade:
In this pull request, we revise the entries surrounding Greengenes and in particular note 2022.10, which is the latest release of Greengenes2.

cc @gregcaporaso

### DIFF
--- a/source/data-resources.rst
+++ b/source/data-resources.rst
@@ -24,6 +24,9 @@ For Silva 138, please cite the following references if you use any of these pre-
 For Greengenes2, please cite:
 
 - McDonald, D. et al. Greengenes2 enables a shared data universe for microbiome studies. bioRxiv (2022). https://doi.org/10.1101/2022.12.19.520774
+
+If using the Naive Bayes classifiers with Greengenes2, please cite:
+
 - Bokulich, N.A., Kaehler, B.D., Rideout, J.R. et al. Optimizing taxonomic classification of marker-gene amplicon sequences with QIIME 2’s q2-feature-classifier plugin. Microbiome 6, 90 (2018). https://doi.org/10.1186/s40168-018-0470-z
 
 Please note, these classifiers were trained using scikit-learn 0.24.1, and therefore can only be used with scikit-learn 0.24.1. If you observe errors related to scikit-learn version mismatches, please ensure you are using the pretrained-classifiers that were published with the release of QIIME 2 you are using.
--- a/source/data-resources.rst
+++ b/source/data-resources.rst
@@ -15,6 +15,8 @@ Naive Bayes classifiers trained on:
 - `Greengenes2 2022.10 full length sequences <http://ftp.microbio.me/greengenes_release/2022.10/2022.10.backbone.full-length.nb.qza>`_ (MD5: ``98d34227fe67b34f62b464466cca4ffa``)
 - `Greengenes2 2022.10 from 515F/806R region of sequences <http://ftp.microbio.me/greengenes_release/2022.10/2022.10.backbone.v4.nb.qza>`_ (MD5: ``43de361005ae6dcae61b078c0c835021``)
 
+.. note:: Greengenes2 has succeeded Greengenes 13_8. If you still need to access the outdated 13_8 classifiers, for example to reproduce old results or to compare against new classifiers, you can access them through the `QIIME 2 2022.11 data resources <https://docs.qiime2.org/2022.11/data-resources/>`_.
+
 For Silva 138, please cite the following references if you use any of these pre-trained classifiers:
 
 - Michael S Robeson II, Devon R O'Rourke, Benjamin D Kaehler, Michal Ziemski, Matthew R Dillon, Jeffrey T Foster, Nicholas A Bokulich. RESCRIPt: Reproducible sequence taxonomy reference database management for the masses. bioRxiv 2020.10.05.326504; doi: https://doi.org/10.1101/2020.10.05.326504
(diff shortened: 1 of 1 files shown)


source/data-resources.rst
```diff
@@ -12,14 +12,22 @@ Naive Bayes classifiers trained on:
 
 - `Silva 138 99% OTUs full-length sequences <https://data.qiime2.org/2022.11/common/silva-138-99-nb-classifier.qza>`_ (MD5: ``b8609f23e9b17bd4a1321a8971303310``)
 - `Silva 138 99% OTUs from 515F/806R region of sequences <https://data.qiime2.org/2022.11/common/silva-138-99-515-806-nb-classifier.qza>`_ (MD5: ``e05afad0fe87542704be96ff483824d4``)
-- `Greengenes 13_8 99% OTUs full-length sequences <https://data.qiime2.org/2022.11/common/gg-13-8-99-nb-classifier.qza>`_ (MD5: ``6bbc9b3f2f9b51d663063a7979dd95f1``)
-- `Greengenes 13_8 99% OTUs from 515F/806R region of sequences <https://data.qiime2.org/2022.11/common/gg-13-8-99-515-806-nb-classifier.qza>`_ (MD5: ``9e82e8969303b3a86ac941ceafeeac86``)
```
gregcaporaso, line 16: Can you include these links as well so users still have the option to access these? I think that will be helpful for a couple of release cycles since gg 13_8 has been such a staple to the field for so long. For example, I've been using GTDB for a while now, but I almost always classify with both GTDB and GG 13_8, and compare assignments for features of interest. 

If you want to emphasize GG2 over 13_8, you could do something like create a box that says something like:

```
GG2 has succeeded Greengenes 13_8. If you still need to access the outdated 13_8 classifiers, for example to reproduce old results or to compare against new classifiers, you can access them through the following links:
...
```

source/data-resources.rst
```diff
@@ -46,14 +54,15 @@ These marker gene reference databases are formatted for use with QIIME 1 and QII
...
 
-Find more information about Greengenes in the `DeSantis (2006) <http://aem.asm.org/content/72/7/5069.full>`_ and `McDonald (2012) <https://www.nature.com/articles/ismej2011139>`_ papers.
+Find more information about Greengenes in the `DeSantis (2006) <http://aem.asm.org/content/72/7/5069.full>`_, `McDonald (2012) <https://www.nature.com/articles/ismej2011139>`_, and `McDonald (2022) <https://www.biorxiv.org/content/10.1101/2022.12.19.520774v1>`_ papers.
 
-**License Information** can be found on the `Greengenes website <https://greengenes.secondgenome.com/>`_. Greengenes data are released under a `Creative Commons Attribution-ShareAlike 3.0 License <https://creativecommons.org/licenses/by-sa/3.0/deed.en_US>`_.
+**License Information** can be found on the `Greengenes website <https://greengenes.secondgenome.com/>`_. Greengenes data (prior to 2022) are released under a `Creative Commons Attribution-ShareAlike 3.0 License <https://creativecommons.org/licenses/by-sa/3.0/deed.en_US>`_. Greengenes data (2022-) are released under a `BSD-3 license <http://ftp.microbio.me/greengenes_release/2022.10/00LICENSE>`_.
```
gregcaporaso, line 65: Should this be _Greengenes2 data (2022-) are released..._? 

Also, a more general question - do you think we should start referring to the older releases collectively as Greengenes1? That's kind of how we handle the QIIME / QIIME 1 / QIIME 2 situation now (where I typically use _QIIME_ to refer to _QIIME 1_ and _QIIME 2_).

gregcaporaso (changes requested):
Thanks @wasade, looking forward to trying this out! Couple of minor requests on this.

source/data-resources.rst
wasade: I don't think so. All Greengenes data (2022-) is under BSD-3. 

I'm torn on that to be honest. With Greengenes, the version numbers are pretty distinct and usually associated with the Greengenes reference. We remarked it as "Greengenes2" to reflect a major change in the database, however (to me at least) the most important thing is still the version number. 

I dunno, what do you think will be least confusing for users?

source/data-resources.rst
gregcaporaso: @wasade, your concern makes sense to me. This transition seems to me to be more akin to the QIIME 1 to QIIME 2 transition rather than Silva release transitions (which tend to be relatively frequent incremental updates). Given that this is a major change to a fundamental tool in the field, I think it's important that we prioritize accessibility of both iterations for a release cycle or two. (For the QIIME transition we went through about a one year period where we supported both to allow users to assess which one they preferred and compare results of the two platforms.) 

While the classifiers can be accessed through old versions of the documentation, I think we can make this more accessible. Would it address your concern if we either:
* added a box with a note like I suggested above, but rather than link users to the classifiers directly, point them to [the relevant section of the 2022.11 version of the documentation](https://docs.qiime2.org/2022.11/data-resources/#taxonomy-classifiers-for-use-with-q2-feature-classifier) 
* or bump the old classifier links to the bottom of this page under a heading of something like _Legacy Taxonomic Classifiers_, and note in that section that we recommend using the most recent versions but that these versions may be useful for reproducing old results or comparing to newer classifiers.

source/data-resources.rst
gregcaporaso: From my perspective calling it Greengenes2 (and then having versions of that, like 2022.11) signifies fundamental changes to the product opposed to an incremental update, and given the long gap between versions I suspect calling it Greengenes2 will be less confusing (and possibly easier to sell to users and/or funders). 

Your call though - just depends on how you want to brand it. I thought it may have been a typo.

source/data-resources.rst
wasade: That makes complete sense, thanks!! I'll issue a commit with the first option

source/data-resources.rst
wasade: Fair point :) I'll revise to Greengenes2 which is consistent with our other uses

gregcaporaso (approved):

gregcaporaso:
Thanks @wasade!

wasade:
Wonderful, thank you!

PR #553, opened 2023-05-08
updates notes about classifiers on data resources page

gregcaporaso:
This PR fixes a broken reference to previous versions of the Greengenes classifiers, and adds a note in several places related to the species-level taxonomy assignments in our classifiers. 

These are two independent commits, so don't need to be squashed before merge.

### DIFF
--- a/source/data-resources.rst
+++ b/source/data-resources.rst
@@ -17,6 +17,8 @@ Naive Bayes classifiers trained on:
 
 .. note:: Greengenes2 has succeeded Greengenes 13_8. If you still need to access the outdated 13_8 classifiers, for example to reproduce old results or to compare against new classifiers, you can access them through the older QIIME 2 data resources pages.
 
+.. note:: The Silva classifiers provided here include species-level taxonomy. While Silva annotations do include species, Silva does not curate the species-level taxonomy so this information may be unreliable. In a future version of QIIME 2 we will no longer include species-level information in our Silva taxonomy classifiers. This is discussed on the QIIME 2 Forum `here <https://forum.qiime2.org/t/processing-filtering-and-evaluating-the-silva-database-and-other-reference-sequence-data-with-rescript/15494#heading--second-header>`_ (see *Species-labels: caveat emptor!*).
+
 For Silva 138, please cite the following references if you use any of these pre-trained classifiers:
 
 - Michael S Robeson II, Devon R O'Rourke, Benjamin D Kaehler, Michal Ziemski, Matthew R Dillon, Jeffrey T Foster, Nicholas A Bokulich. RESCRIPt: Reproducible sequence taxonomy reference database management for the masses. bioRxiv 2020.10.05.326504; doi: https://doi.org/10.1101/2020.10.05.326504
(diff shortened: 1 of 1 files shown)


source/data-resources.rst
```diff
@@ -15,7 +15,9 @@ Naive Bayes classifiers trained on:
...
 
-.. note:: Greengenes2 has succeeded Greengenes 13_8. If you still need to access the outdated 13_8 classifiers, for example to reproduce old results or to compare against new classifiers, you can access them through the `QIIME 2 2023.5 data resources <https://docs.qiime2.org/2023.5/data-resources/>`_.
+.. note:: Greengenes2 has succeeded Greengenes 13_8. If you still need to access the outdated 13_8 classifiers, for example to reproduce old results or to compare against new classifiers, you can access them through the older QIIME 2 data resources pages.
+
+.. note:: The Silva classifiers provided here include species-level taxonomy. While Silva annotations do include species, Silva does not curate the species-level taxonomy so this information may be unreliable. In a future version of QIIME 2 we will no longer include species-level information in our Silva taxonomy classifiers. This is discussed on the QIIME 2 Forum `here <https://forum.qiime2.org/t/processing-filtering-and-evaluating-the-silva-database-and-other-reference-sequence-data-with-rescript/15494#heading--second-header>`_ (see *Species-labels: caveat emptor!*).
```
nbokulich, line 20: the info is indeed unreliable — if I recall correctly, we found that something like 75% of the species labels do not match the genus! (ref: rescript)

but removing this info might disappoint many users who have come to expect this. We could release versions with/out the species annotations. What do you think, @mikerobeson ?

gregcaporaso: I feel like if the 75% of the species labels are unreliable this is going to be misleading to a lot of folks (this is actually how it came to my attention - it was pointed out to a colleague by a reviewer when they were talking about Silva species labels in a manuscript that was under review). So I would be in favor of just not including these, but providing details (e.g., as in the current forum post) on how to train a classifier that does include them.

gregcaporaso: Also, sounds like there are no objections to adding this note to the page in the docs?

colinbrislawn: I have a quick question

When parent taxonomy diverg, matching s__ levels don't make much sense, i.e.
g__Serratia; s__Oryza_sativa
g__Hemiptera; s__Oryza_sativa

In this new system, what happens when parent taxonomy and the first two words of s__ agree, but the following subspecies don't match?
s__Clostridioides_difficile
s__Clostridioides_difficile_R20291

Are these kept as two different s__Clostridioides_difficile?
Are these merged into a single s__Clostridioides_difficile?
Are we dropping s__ totally?

mikerobeson: RE s__Clostridioides_difficile: as explained in the tutorial the first two words are kept. Nothing is "merged" unless the user picks one of the other dereplication modes in rescript, *i.e.* they might choose to return the LCA taxonomy, and/or cluster the sequences. By default, the `uniq` mode of `rescript dereplicate`, will retain all unique sequences, as well as any sequences that are identical but have differing taxonomy. We essentially leave the decision up to the classifier. This also allows us to use weights. That is,  to allow the classifier to have a chance of choosing one identical sequence over another, *i.e.* q2-clawback.

RE s__Oryza_sativa: yep, they do not make sense. In this case the source organism (*Oryza sativa*), is from which the microbial sequence was derived. Hence the taxonomy from domain down through genus, is the microbial sequence, but the "species label" is from host specific epithet.

gregcaporaso:
@nbokulich, @mikerobeson - it sounds like we're in agreement on this note and the path forward. 

This one should be ready for merge.

PR #561, opened 2023-10-06
NEW: distro updates & removing gneiss

lizgehret:


ebolyen (changes requested):
The tabs for each distro don't work because they all share the same `id=""` html attrs.

We should either make every ID and href=#id unique, or use classes and update the javascript to tab all distros at once. Either seems fine with me.

ebolyen (approved):
Fixed

PR #568, opened 2023-11-16
MAINT: remove mamba from install docs

lizgehret:
removes mamba from the QIIME 2 install docs now that conda's primary solver uses libmamba.


lizgehret:
Locally rendered updated install instructions, removing `mamba install` commands.

Screen Shot 2023-11-16 at 11 12 09 AM

PR #575, opened 2024-02-20
BUG: adding vsearch merge-pairs API change

lizgehret:
`vsearch merge-pairs` API was changed [here](https://github.com/qiime2/q2-vsearch/pull/95), which also needs to be updated in the docs commands to prevent doc build failures.


lizgehret:
**Note**: I reverted the dev-bump for this change as well so that changes from this branch can be used to rebuild the docs successfully for 2024.2.

PR #578, opened 2024-05-16
Remove classifier text and replace with external link

Oddant1:
Replace the classifier stuff in the docs with a link to https://resources.qiime2.org

### DIFF
--- a/source/data-resources.rst
+++ b/source/data-resources.rst
@@ -8,8 +8,6 @@ Taxonomy classifiers have been moved to the external page https://resources.qiim
 
 .. _`marker gene db`:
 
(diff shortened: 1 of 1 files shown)


Oddant1:
Do we want that note there or not? The way the code is flowed suggests it's supposed to be associated with the `Marker gene reference databases` section, but it looks awkward, and there is already another copy of that note at the bottom of that section.

image

Oddant1:
Looks like this without the note, I think without is much better I only ask because like I said the way the code is written looks like the intent is for the note to be associated with the section beneath it 
image

gregcaporaso:
@Oddant1, we don't need that note, so let's go with your second example above. I suspect that note was just misplaced with respect to the anchor that follows it right now.

PR #582, opened 2024-06-05
MAINT: rename shotgun->metagenome and update plugin list

nbokulich:
* renames shotgun distro --> metagenome
* updates plugin list (for both amplicon and metagenome distros), as this was out of date.


lizgehret:
going to merge this now that it's passed - I'll rebuild locally this evening (currently on a plane with slow wifi) and then get these uploaded to the server by tomorrow morning!

PR #586, opened 2024-06-25
IMP: install instructions use direct URL for env file

lizgehret:
URLs can now be used in the `-f` param of `conda env create`. Replacing a local file download with a direct raw GH link should reduce the amount of install errors users run into by:

- removing the need to use `wget`
- removing the possibility that users download differing files with the same name (i.e. if a release env gets patched, the file name remains the same) and then aren't using the correct file for install

This adjustment reduces the number of steps required for install and creates a more concise process.

### DIFF
--- a/source/install/native.rst
+++ b/source/install/native.rst
@@ -70,32 +70,20 @@ QIIME 2 Amplicon Distribution
             </p>
          </div>
          <div id="amplicon-macOS-intel" class="tab-pane fade">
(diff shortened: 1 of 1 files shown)


colinvwood (approved):

colinvwood:
Looks good and the installation worked for me.

lizgehret:
For the changelog:

> @lizgehret updated the installation instructions to now use a direct URL for the environment file. This removes the need to install `wget` or download the file to your local machine!

PR #593, opened 2024-10-30
DEV-BUMP: 2025.4

lizgehret:
~~**NOTE**: the python version will need to change if we update to 3.11 in the next release.~~ disregard, python version isn't slated for update in 2025.4


lizgehret:
blocked by #594 (will cause merge conflicts, so waiting until that's merged to address here)

